PR #79, opened 2020-12-14
Dev/andris/makefile tools/testing

andreeis:


src/logger.ts
```diff
@@ -30,20 +30,15 @@ function loggingLevelApplies(messageVerbosity: string | undefined): boolean {
 }
 
 //TODO: implement more verbosity levels for the output log
-export function message(message: string, loggingLevel?: string): void {
+export function message(message: string, loggingLevel?: string, show: boolean = true): void {
```
bobbrow, line 33: My recommendation here is to make "show" be event-based instead of message based.  For example, you show the window when you begin a build.  Or you show the window when the output channel is first created.  Then you leave it to the user to decide when they want to see the output.

src/parser.ts
```diff
@@ -617,6 +639,19 @@ function parseFilesFromToolArguments(args: string, exts: string[]): string[] {
     match = regexp.exec(args);
     while (match) {
         let result: string = match[1];
+
+        // It is quite common to encounter the following pattern:
+        //  `test -f 'sourceFile.c' || echo './'`sourceFile.c
+        // Until we implement the correct approach (to query live the test command)
+        // we can just ignore it and consider the second option of the OR
+        // (by removing './'`). Even if this is hacky, it is worth it
```
bobbrow, line 647: ok. can we track this with an issue?

src/parser.ts
```diff
@@ -159,23 +149,41 @@ export async function preprocessDryRunOutput(cancel: vscode.CancellationToken, d
...
+    // Line with the original command, containing a 'test' command to determine on the fly the source code path.
+    // This line is present in the dryrun and also in the build log. Can't easily parse the correct source code path.
+    // /bin/bash ./libtool  --tag=CC   --mode=compile gcc -DHAVE_CONFIG_H   -I./include -I./include  -DGC_PTHREAD_START_STANDALONE    -fexceptions -Wall -Wextra -Wpedantic -Wno-long-long -g -O2 -fno-strict-aliasing  -MT cord/libcord_la-cordxtra.lo -MD -MP -MF cord/.deps/libcord_la-cordxtra.Tpo -c -o cord/libcord_la-cordxtra.lo `test -f 'cord/cordxtra.c' || echo './'`cord/cordxtra.c
+    // Line with the resolved command, from which the extension can parse a valid source code path.
+    // This line is present only in the build log, immediately following the above line.
+    // libtool: compile:  gcc -DHAVE_CONFIG_H -I./include -I./include -DGC_PTHREAD_START_STANDALONE -fexceptions -Wall -Wextra -Wpedantic -Wno-long-long -g -O2 -fno-strict-aliasing -MT cord/libcord_la-cordxtra.lo -MD -MP -MF cord/.deps/libcord_la-cordxtra.Tpo -c cord/cordxtra.c  -fPIC -DPIC -o cord/.libs/libcord_la-cordxtra.o
+        preprocessTasks.push(function () {
```
bobbrow, line 164: nit: unindent

src/parser.ts
```diff
@@ -159,23 +149,41 @@ export async function preprocessDryRunOutput(cancel: vscode.CancellationToken, d
...
+        regexp = /libtool: compile:|libtool: link:/mg;
+        preprocessedDryRunOutputStr = preprocessedDryRunOutputStr.replace(regexp, "\nLIBTOOL_PATTERN\n");
+    });
+
     // Process some more makefile output weirdness
     // When --mode=compile or --mode-link are present in a line, we can ignore anything that is before
```
bobbrow, line 170: >mode-link [](start = 32, length = 9)

update comment `--mode=link`

src/parser.ts
```diff
@@ -159,23 +149,41 @@ export async function preprocessDryRunOutput(cancel: vscode.CancellationToken, d
...
 
-    // Remove lines with $ since they come from unexpanded yet variables. The extension can't do anything yet
+    // Remove lines with $() since they come from unexpanded yet variables. The extension can't do anything yet
     // about them anyway and also there will be a correspondent line in the dryrun with these variables expanded.
+    // Don't remove lines with $ without paranthesis, there are valid compilation lines that would be ignored otherwise.
     preprocessTasks.push(function () {
-      regexp = /.*\$.*/mg;
-      preprocessedDryRunOutputStr = preprocessedDryRunOutputStr.replace(regexp, "");
+      regexp = /.*\$\(.*/mg;
+      preprocessedDryRunOutputStr = preprocessedDryRunOutputStr.replace(regexp, "SomethingGood");
```
bobbrow, line 183: >SomethingGood [](start = 81, length = 13)

Not sure if this is correct :)

src/cpptools.ts
```diff
@@ -114,32 +114,33 @@ export class CppConfigurationProvider implements cpp.CustomConfigurationProvider
...
         }
-        logger.message("----------------------------------------------------------------------------");
+        logger.message("----------------------------------------------------------------------------", "Verbose");
     }
 
     public logConfigurationProviderItem(filePath: cpp.SourceFileConfigurationItem, fromCache: boolean = false): void {
-        logger.message("Sending configuration " + (fromCache ? "(from cache) " : "") + "for file " + filePath.uri.toString() + " -----------------------------------", "Verbose");
-        logger.message("    Defines: " + filePath.configuration.defines.join(";"), "Verbose");
-        logger.message("    Includes: " + filePath.configuration.includePath.join(";"), "Verbose");
+        let uriObj: any = filePath.uri;
```
bobbrow, line 129: >filePath.uri [](start = 26, length = 12)

Since we own this data, we can assert that it is `vscode.Uri`
```
  let uriObj: vscode.Uri = <vscode.Uri>filePath.uri;
```

Or you can just leave it as `any` and do uriObj.path below.

src/util.ts
```diff
@@ -317,6 +317,30 @@ export function removeQuotes(str: string): string {
     return str;
 }
 
+// Remove only he quotes that are surrounding the given string.
+export function removeSurroundingQuotes(str: string): string {
+    str = str.trim();
+    if (str.startsWith('"') && str.endsWith('"')) {
+        str = str.substring(1, str.length - 1);
```
bobbrow, line 324: >str.length - 1 [](start = 31, length = 14)

Can you double check that this is supposed to be an index instead of a length?

src/util.ts
```diff
@@ -317,6 +317,30 @@ export function removeQuotes(str: string): string {
...
+
+    return str;
+}
+
+// Escape every character of the given string.
+// Used when constructing a regular expression from file names which can contain
+// special characters.
+// To avoid more complicated character analysis, we can escape all characters
+// because escaping a non special character is a no-op.
+export function escapeString(str: string): string {
```
bobbrow, line 335: Can you try this instead?  It comes from cpptools:

```
const escapeChars: RegExp = /[\\\^\$\*\+\?\{\}\(\)\.\!\=\|\[\]\ \/]/;  // characters that should be escaped.

// Insert '\\' in front of regexp escape chars.
function escape(chars: string): string {
    let result: string = "";
    for (const char of chars) {
        if (char.match(escapeChars)) {
            result += `\\${char}`;
        } else {
            result += char;
        }
    }
    return result;
}
```

src/parser.ts
```diff
@@ -617,8 +661,27 @@ function parseFilesFromToolArguments(args: string, exts: string[]): string[] {
...
+            result = result.substring(5, result.length);
+        }
+
         if (result) {
-            result = result.trim();
+            result = util.removeSurroundingQuotes(result);
+
+            // Debug message to identify easier the scenarios where source files have inner quotes.
+            if (result.includes('"')) {
+                logger.message(`File argument ${result} contains quotes.`, "Debug");
```
bobbrow, line 682: >${result} [](start = 46, length = 9)

maybe add `\`` to surround the output?

src/parser.ts
andreeis: https://github.com/microsoft/vscode-makefile-tools/issues/81

src/logger.ts
andreeis: Yes, good idea. Will do in a separate PR.

src/util.ts
andreeis: Yes, index.

PR #100, opened 2021-02-10
Add build and launch commands

andreeis:
To be able to hook a build before debug/run via launch.json, we add 2 additional launch commands (makefile.buildAndLaunchTargetPath, makefile.buildAndLaunchTargetFileName), from the makefile.launchTargetPath family.  Invoking these is also triggering a build of the current build target, if setting makefile.buildBeforeLaunch allows.
There wasn't a need to add makefile.buildAndLaunch... for args and directory.

Very important to note that launch.json should only have one makefile.buildAnd... command defined, otherwise more builds will be spawned at the same time. Makefile Tools extension doesn't have an easy way to control this since VSCode is calling into our commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -417,12 +417,30 @@
                     "scope": "resource"
                 },
                 "makefile.phonyOnlyTargets": {
-                  "type": "boolean",
-                  "default": false,
-                  "description": "Display only the phony targets",
-                  "scope": "resource"
-              }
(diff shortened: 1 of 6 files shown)


andreeis:
Closing this PR and opening a new one (including the suggestions that we discussed) fresh from main.

Issue #101, opened 2021-02-10
Implement support for VSCode tasks


andreeis:
The most urgent reason for tasks support is the ability to build before debug/run when launch.json is used.
Implementing this will make the commands makefile.launchTargetPath and makefile.launchTargetFileName not useful anymore and we will be able to remove them.

ericis:
@andreeis I always define `.PHONY` command orchestrations inside the "./makefile" itself, which handles all of the command dependencies. Would this work for your scenario? This has the added benefit that any engineer that can run make can execute the tasks as defined in the "./makefile" without requiring VS Code.

jankap:
Any updates on this? :)

andreeis:
@jankap, we were not able to get to this work item yet. This is currently also not part of the upcoming 0.4 release. We would gladly review and test a PR (to be included in a sooner release) if anyone in the community is interested to implement this support before we do.

MightyCreak:
I would add that when setting up a toolchain in VSCode, you expect to have new tasks and new launch targets in the tasks.json and launch.json (respectively) of the workspace.

I just tried Makefile Tools and it feels like it is using is own solution instead of going with the VSCode's workflow. For instance, if Makefile Tools was using tasks.json and launch,json, it would be very easy for the user to add arguments to any of the tasks or targets.

gcampbell-msft:
@MightyCreak Could you describe in more detail the scenario and workflow that you're expecting? There shouldn't be anything in the makefile extension that stops you from using tasks.json and launch.json, but are you asking for more support in the Makefile Tools extension regarding this? 

It'd be great to get a better understanding of your desired scenario. Thanks!

MightyCreak:
Keep in mind that, I have a good knowledge of Visual Studio, but my knowledge of VSCode is not that great 😉

From what I understand, the classic way to debug is to hit F5 in VSCode, and to do so you need to setup a "target" in `launch.json`, and in this target you have a `preLaunchTask` properties set to the build task form `tasks.json`.

And this is the part where VSCode is a bit confusing to me. It seems that for some languages, VSCode is able to autogenerate these two files if they don't exist and there is not a lot to do to make it work perfectly.

What I'd expect from Makefile Tools when configuring a new project, would be to add these target and task automatically (if it's possible). So that when I hit F5, it runs `make`, and if the compilation succeeds, it runs the binary and attach the debugger to it.

Does it make sense?

gcampbell-msft:
[...]

PR #102, opened 2021-02-10
Launch commands renaming and ability to trigger build from launch.json

andreeis:
Renamed the launch commands that existed already with a get prefix. These do not trigger a build.
Added 2 new counterparts, without the get prefix, that trigger a build if makefile.buildBeforeLaunch allows. These are solving the problem of building before debug/run when launching is invoked via launch.json.

makefile.launchTargetPath and makefile.launchTargetFileName are temporary commands that are only useful for invocations written in launch.json. 
Throughout the whole launch.json, there should be used only one of these two new commands and only once, otherwise a separate build will be spawned for each invocation. The extension can't know how many times VSCode will call into it with any number of these variables.

makefile.getLaunch.... family of commands can also be called from launch.json but they will not trigger a build.

The correct flow for building before launching is via the preLaunchTask (keyword part of the launch.json schema), for which Makefile Tools needs to implement tasks support. We will do this in the immediate future: https://github.com/microsoft/vscode-makefile-tools/issues/101.


src/extension.ts
```diff
@@ -222,29 +222,39 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         await launcher.runCurrentTarget();
     }));
 
-    context.subscriptions.push(vscode.commands.registerCommand('makefile.launchTargetPath', () => {
-        telemetry.logEvent("launchTargetPath");
-        return launcher.launchTargetPath();
+    context.subscriptions.push(vscode.commands.registerCommand('makefile.getLaunchTargetPah', () => {
```
bobbrow, line 225: Path is misspelled "Pah" in the next few lines

src/launch.ts
```diff
@@ -157,9 +179,9 @@ export class Launcher implements vscode.Disposable {
             type: dbg,
             name: `Debug My Program`,
             request: 'launch',
-            cwd: '${command:makefile.launchTargetDirectory}',
+            cwd: '${command:makefile.getLaunchTargetDirectory}',
             args,
-            program: '${command:makefile.launchTargetPath}',
+            program: '${command:makefile.getLaunchTargetPah}',
```
bobbrow, line 184: Would we rather have it build before debugging by default?

bobbrow (changes requested):
:clock1:

src/launch.ts
andreeis: This is the internal launch functionality and there is already a build before this point, also controlled by the same setting.

src/extension.ts
andreeis: I fixed these errors but didn't realized that I didn't push an update commit. Thanks.

src/launch.ts
bobbrow: So since this is the just the internal launch functionality, should we call `getLaunchTargetPath()` ourselves and put the expected string in here instead of sending a command variable which will result in another callback to resolve it?

andreeis: Oh interesting, I didn't realize that. Good point.

bobbrow (approved):

Issue #104, opened 2021-02-10
Feature request: possibility to generate compile_commands.json


rapgenic:
The [compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html) file is quite useful for many tools in the C/C++ ecosystem, and given that the extension already collects all the required information it should be easy to write down that file on configuration completed.

This would also allow to get rid of other software like bear or compiledb, that are used to generate such file, which might be tricky to install.

I am working on the code right now, and I already have a working proof of concept; if you are interested I can open a pull request for you to review the code (it's a quite simple modification)

andreeis:
Thank you for the feature idea and for willing to help us implement it. Indeed this would be a very useful functionality.
We are interested in reviewing and testing a PR from you. We would merge it in after we ship Makefile Tools on the MS Market Place next week. You and any other developers in the community will be able to use it immediately even if it won't be included in our first public release because we can build a VSIX after every PR is completed.
Looking forward to see your code and incorporate it in the extension.

andreeis:
@rapgenic, we released Makefile Tools 0.3.0 today which includes this feature you implemented for us. Thank you very much for your contribution. Let us know if you encounter any other issues.

If you are happy with the extension, you can leave us a review on [Market Place](https://marketplace.visualstudio.com/items?itemName=ms-vscode.makefile-tools&ssr=false#review-details).

PR #107, opened 2021-02-11
[WIP] Add compile_commands.json generator

rapgenic:
This PR contains what I've done so far for issue #104.

It adds two options, similar to cmake tools':
 * `exportCompileCommandsFile`: a boolean to toggle the generation of the compilation database
 * `compileCommandsPath`: the destination path. Note that this is a little different from cmake tools, because it does not support placeholders (e.g. `${workspaceFolder}`, it seems to me that they're not supported right now), and the path is automatically considered as a subpath of the workspace folder (which in my opinion might be better than cmake tools, which requires to prepend the path with an often obvious `${workspaceFolder}`).

In the `parser.ts` file I added a function `parseCompileCommands` modeled after `parseCustomConfigProvider` that extracts from the dry run log the compilation info and for each file creates a `CompileCommand` entry.

On configuration completed if the option is enabled, the `compile_commands.json` file is automatically written/overwritten if present.

I've tested it with my projects, but I'm not sure if it works in any case, so I'd be happy to hear your review and fix any issues you have with my implementation, or code architecture, since I structured the patch mainly by imitation of surrounding code...

### DIFF
--- a/package.json
+++ b/package.json
@@ -439,6 +439,17 @@
                     "default": true,
                     "description": "Clear the output channel at the beginning of a build",
                     "scope": "resource"
+                },
+                "makefile.exportCompileCommandsFile": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Enable generation and exporting of compile_commands.json"
+                },
+                "makefile.compileCommandsPath": {
+                    "type": "string",
+                    "default": "compile_commands.json",
+                    "description": "The path to the compilation database file",
+                    "scope": "resource"
                 }
             }
         },
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -1139,9 +1139,6 @@ export async function initFromStateAndSettings(): Promise<void> {
             let updatedExportCompileCommandsFile: boolean | undefined = workspaceConfiguration.get<boolean>(subKey);
             if (updatedExportCompileCommandsFile !== exportCompileCommandsFile) {
                 readExportCompileCommandsFile();
-                // This will trigger a reread from the filesystem of the existing
-                // compile_commands.json
-                make.clearCompileCommands();
                 updatedSettingsSubkeys.push(subKey);
             }
 
@@ -1152,9 +1149,6 @@ export async function initFromStateAndSettings(): Promise<void> {
             }
             if (updatedCompileCommandsPath !== compileCommandsPath) {
                 readCompileCommandsPath();
-                // This will trigger a reread from the filesystem of the existing
-                // compile_commands.json
-                make.clearCompileCommands();
                 updatedSettingsSubkeys.push(subKey);
             }
 
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -275,6 +275,29 @@ export function readConfigurationCachePath(): void {
     logger.message(`Configurations cached at ${configurationCachePath}`);
 }
 
+let compileCommandsPath: string | undefined;
+export function getCompileCommandsPath(): string | undefined { return compileCommandsPath; }
+export function setCompileCommandsPath(path: string): void { compileCommandsPath = path; }
+export function readCompileCommandsPath(): void {
+    let workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("makefile");
+
+    compileCommandsPath = workspaceConfiguration.get<string>("compileCommandsPath");
+    if (compileCommandsPath) {
+        compileCommandsPath = util.resolvePathToRoot(compileCommandsPath);
+    }
+
+    logger.message(`compile_commands.json path: ${compileCommandsPath}`);
+}
+
+let exportCompileCommandsFile: boolean | undefined;
+export function getExportCompileCommandsFile(): boolean | undefined { return exportCompileCommandsFile; }
+export function setExportCompileCommandsFile(value: boolean): void { exportCompileCommandsFile = value; }
+export function readExportCompileCommandsFile(): void {
+    let workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("makefile");
+    exportCompileCommandsFile = workspaceConfiguration.get<boolean>("exportCompileCommandsFile");
+    logger.message(`Export compile_commands.json: ${exportCompileCommandsFile}`);
+}
+
 let additionalCompilerNames: string[] | undefined;
 export function getAdditionalCompilerNames(): string[] | undefined { return additionalCompilerNames; }
 export function setAdditionalCompilerNames(compilerNames: string[]): void { additionalCompilerNames = compilerNames; }
@@ -785,6 +808,8 @@ export async function initFromStateAndSettings(): Promise<void> {
     readBuildBeforeLaunch();
     readClearOutputBeforeBuild();
     readIgnoreDirectoryCommands();
+    readExportCompileCommandsFile();
+    readCompileCommandsPath();
 
     analyzeConfigureParams();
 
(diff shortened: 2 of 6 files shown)


rapgenic:
Possible improvements:
* Reconfigure after changing the settings
* Telemetry (this is a bit obscure to me, I would need some pointers)
* Deleting the old compile_commands.json on path change
* more?

andreeis:
Thank you for the PR. I will review in more detail later and also brainstorm on the improvements ideas.
Was there any reason you had to add a new configure sub-phase for the generation of compile_commands.json? I wonder if you can hook up the creation of CompileCommand entries directly from parseCustomConfigProvider, to benefit of that same looping through output lines and of how that function already parses everything.

rapgenic:
Thank you for your quick review and support!

> Was there any reason you had to add a new configure sub-phase for the generation of compile_commands.json? I wonder if you can hook up the creation of CompileCommand entries directly from parseCustomConfigProvider

Absolutely right, I had done it for code separation, but it isn't really necessary as you pointed.

rapgenic:
I have discovered that I had completely missed the possibility to have a non clean configuration, in which case the compile_commands.json file ends up being empty. Probably you'll want to wait for me to fix that before doing a full review...

andreeis:
Not done reviewing yet. I just want to share some thoughts as soon as they cross my mind. Don't rewrite anything yet, let's figure out what's best first.
Anyway it looks good in general :), thank you. And it's normal you wouldn't know of the below facts and assumptions, which may cause some rewrites to be needed and I hope that's not too frustrating.

1. There is some overlap between the functionality you implemented around reading and merging with a previous compile_commands.json and the extension configuration cache functionality:
    - a clean configure writes a new .vscode/configurationCache.log
    - a non clean configure reads from .vscode/configurationCache.log and appends anything extra
    - at the end of configure phase, even if we have a non clean configure operation (and a valid cache was created previously) we still have all data in memory.
        - So, I'm thinking that maybe you can create the compile commands entries also at the very end (when you write the file) and read from the final result of the configure process instead.
        - this way, the new implementation would interact less with what's already there (no extra callback, no merging with a previous compile_commands.json, no unnecessary interference with the parsing of the log lines, no forcing of clean configure which can cause problems, see below bullet, ...)

2. Problems with triggering a clean configure when a non clean configure was requested:
    - for some repositories, a clean configure may last for hours (simply because that's how much dry-run takes, depends how the makefile is written)
    - we also avoided as much as possible to decide 'under the covers' (for the user) to do a clean configure instead of a non clean one. The only time the extension does this is when the project is loaded and if we have a cache, it stays unblocking in the background. You can also change your call to setConfigureIsInBackground from false to true for the same reason
    - the users have options in place to disable automatic re-configure (makefile.configureOnCommand or configureOnEdit set to false). You can also check for that before changing a non clean configure to a clean configure. 
    - but even if those settings don't signal explicitly that the user wants to void unexpected automatic configuring and even if you make the clean configure not blocking and in the background, maybe it's still not worth to add the complexity and risk of this approach (forcing clean configure over non clean configure) when we know (from what I explained in bullet 1) that at the end of configure we have all in memory anyway.

rapgenic:
Thank you for the support, and I'll be more than happy to rewrite anything if necessary!

> So, I'm thinking that maybe you can create the compile commands entries also at the very end (when you write the file) and read from the final result of the configure process instead.

Please correct me if i'm wrong, I understand that what you're suggesting is to use the information stored in the in memory configuration cache, to generate the compilation database. This was actually my first approach, which however I discarded because I didn't like the idea to rebuild the compilation command from the already processed data (might come out somewhat different). However another possibility might be to store in the cache the entire command as well. This way we wouldn't need to alter any preexisting mechanism, besides adding an entry to the cache (even if a bit redundant). What do you think about that?

andreeis:
Oh, I missed the fact that the original command itself was missing. Reconstructing the command from the pieces we parse instead of providing the original form is not a good idea, you are right. Including it as a separate field has some redundancy and I can see how you went the other way first. I think a separate field is less risky than forcing a clean configure, but let me bake a few hours on that and I'll get back to you.

andreeis:
Just to make sure I don't overlook any downside of the extra configuration cache field approach. When you considered it, what reasons did you have to opt for caching the compile_commands format instead, besides the redundancy?

rapgenic:
Basically the main downsides of adding a extra cache field are:
* The cache works with and interfaces with objects that come from the cpptools API, so adding a field for the command might mean modifying the cpptools api (and maybe the extension as well, as it seems that the content to cache is read directly from the cpptools provider?)
* In alternative we might add a new map, separated from the cpptools provider, to store only the commands, and given that we'd probably need to store the `directory` field as well (the path where the command is executed, which I'd forgot about before), we would end up storing in the cache the entire compilation commands database, which is more or less the same as using the compile_commands.json file itself as a cache.
* The only advantage of storing the compilation database in the cache and then writing it in the filesystem might be that we could reuse the existing cache loading/saving infrastructure and that if the compile_commands.json were to be deleted/modified by external tools it could be recreated easily without requiring a clean build, being the data cached.

rapgenic:
Quick update, I've experimented with the first option (extending the cpptools api) and storing in the fileindex an additional compileCommand field. From quick testing everything seems to work, for now I'll push my new commit in a new branch on my fork, in case you're still reviewing this PR (the branch name is compilecommandscache)

rapgenic:
One problem I did not initially consider is that for a single file more than one compilation command can be specified (if they are compiled multiple times) and can be distinguished via the `output` field. This means that in some (edge) cases the current file based indexing/merging approach will not work, as, to be certain not to overwrite a different compilation command entry, we should index them by the source *and output* file. This is quite an edge case, but if we want to fully support the compile_commands.json specification, we might need to find a different approach.

andreeis:
Thank you for the excellent points and the very thorough analysis. We had a long weekend in the US and that is why this reply is late.
I'll confirm more in about a day. I am leaning towards implementing first bullet now (as you said you already did and tested) and I am also tempted by replacing completely the current cache with the compile commands file, but later in the future since this is more a refactoring/improvement of the source code than new functionality.

Regarding multiple and different compilations for the same source file, I always imagined that the user would know to select a particular build target instead of "all". For example, if I write a makefile compiling the same source.cpp for 3 different target architectures, if I build "all" I would get all 3 of them built but then if I develop in the code base and I want relevant IntelliSense I would select ArchitectureX as a target, which would cause only one compilation command to be detected for source.cpp. So, how is compile_commands.json handling this scenario? More entries of the same source file but with different command (since some switches would differ and of course the output path)?

rapgenic:
> I am also tempted by replacing completely the current cache with the compile commands file, but later in the future since this is more a refactoring/improvement of the source code than new functionality.

I did not think about that, but that might be a great option in my opinion, given that right now you extract all the information from the dryrun/build log, and the compile_commands.json is meant to be a more machine readable version of the build log.

> So, how is compile_commands.json handling this scenario? More entries of the same source file but with different command (since some switches would differ and of course the output path)?

I think that this mostly boils down to how every tool interprets the compile_commands.json: from what I understand, the format itself is just a log of the commands, associated to a source file and, if necessary to distinguish between different builds of the same file, an output file (which means that there might be more entries for the same file, even though I'd consider the compilation database more of a list of n-uples than a map from file to compilation command). 

In the end I guess it's every tool responsibility to detect that a file has multiple compilation commands and choose the one that fits for its purpose (for example I imagine some tools might even want to read the compilation database iterating the outputs rather than the source files). 

However I think this is some really edge case (I have never encountered it before), and for now might not be worth overcomplicating the implementation. Plus we would get it for free if you intend to replace the current cache with the compile_commands.json file itself in the future.

andreeis:
Agree to worry about different compilations for the same source file later.
Opened work item to replace the current configuration cache with the compilation database: https://github.com/microsoft/vscode-makefile-tools/issues/110.
I reviewed your branch "compilecommandscache" and it looks good. I also would like to clone it and test for more repositories that we used to stabilize the extension but I will get to this in a few days.

rapgenic:
Hi, sorry for the late reply, I was a bit busy the last two days..

> I reviewed your branch "compilecommandscache" and it looks good. I also would like to clone it and test for more repositories that we used to stabilize the extension but I will get to this in a few days.

That's great, I will merge it here so that you can better review it!

andreeis:
@rapgenic, we released 0.2.1 and we are preparing for 0.3 which should be ready in 3-4 months if all goes according to plan.
We didn't want to include this PR in 0.2 since we focused only on bug fixes and we didn't want to destabilize the extension with new features but now is a good time and I'm validating your PR again to merge it in.

rapgenic:
@andreeis thank you for the update, I'm really looking forward to it. Please tell me if there is something new to fix!

PR #112, opened 2021-02-17
update package version and readme

bobbrow:
@andreeis I think we need to document the command variable substitutions for customers who prefer to use `launch.json` instead of the debug/run buttons.  Something similar to this document: https://github.com/microsoft/vscode-cmake-tools/blob/develop/docs/cmake-settings.md#command-substitution

Could you make a page like this and add it to the docs folder?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,9 @@
-# VS Code Makefile Tools - early preview
+# VS Code Makefile Tools
 
 This extension provides IntelliSense configurations to the VS Code C/C++ Extension for Makefile projects.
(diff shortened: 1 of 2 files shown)


andreeis:
Yes I will create a page similar to that.
I approve this PR.
I realized the readme needs also a paragraph about the pre-configure command. This allows a script to be run (and the environment persisted for further operations) before configuring.
I'll include this in my next PR that I'm starting now.

andreeis (approved):

PR #113, opened 2021-02-17
Update readme. Add settings doc.

andreeis:
Updated the readme with "pre-configure" and "configuring from log" information.
Added a new document describing the settings and commands, following the CMake Tools example.
I did not list all the settings in package.json (with their name, description and default values). Let me know if you think I should add that now. The CMT document was listing a subset.

### DIFF
--- a/docs/makefile-settings.md
+++ b/docs/makefile-settings.md
@@ -1,0 +1,30 @@
+# Configure Makefile Tools settings
+
+Makefile Tools supports a variety of settings that can be set at the user, or workspace, level via VSCode's `settings.json` file. This topic covers the available options and how they are used.
+
+## Makefile settings
+
+There is no variable substitution available for the Makefile Tools settings and we plan to implement such support in a future release.
+
+### Command substitution
+
+Makefile Tools can expand VS Code commands. For example, you can expand the path to the launch target by using the syntax `${command:makefile.launchTargetPath}`
+
+Be careful with long-running commands because it isn't specified when, or how many times, Makefile Tools will execute a command for a given expansion.
+
+Supported commands for substitution:
+
+|command|substitution|
+|-------|------------|
+|`makefile.getLaunchTargetPath`|The full path to the target executable, including the filename. The existence of the target is not validated.|
+|`makefile.getLaunchTargetDirectory`|The full path to the target executable's directory. The existence of the directory is not validated.|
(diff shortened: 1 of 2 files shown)


docs/makefile-settings.md
```diff
@@ -0,0 +1,30 @@
+# Configure Makefile Tools settings
+
+Makefile Tools supports a variety of settings that can be set at the user, or workspace, level via VSCode's `settings.json` file. This topic covers the available options and how they are used.
```
bobbrow, line 3: I would say `tasks.json` and `launch.json` instead of `settings.json`

bobbrow: Really, we just need the Command Substitution section and an explanation that this is for tasks/launch json files.  We can consider whether it makes sense to add the rest of the sections later.

andreeis: Done. Simplified to include only command expansion but I also mentioned that variable substitution will be implemented at a later time, so users know this in advance. There were questions about that.

docs/makefile-settings.md
```diff
@@ -0,0 +1,19 @@
...
+|`makefile.getLaunchTargetDirectory`|The full path to the target executable's directory. The existence of the directory is not validated.|
+|`makefile.getLaunchTargetFileName`|The name of the target executable file without any path or extension information. The existence of the target is not validated.|
+|`makefile.getLaunchTargetArgs`|A string array describing the arguments of the current launch target.|
+|`makefile.getLaunchTargetArgsConcat`|A string describing the arguments of the current launch target concatenated in a single string and separated by space.|
+|`makefile.launchTargetPath`|The full path to the target executable, including the filename. If `makefile.buildBeforeRun` is true, invoking this substitution will also start a build.|
+|`makefile.launchTargetFileName`|The name of the target executable file without any path or extension information. If `makefile.buildBeforeRun` is true, invoking this substitution will also start a build.|
+
+# Variable substitution
+
+Variable substitution is not yet implemented and it will be supported in a future release.
```
bobbrow, line 19: Do we plan to add any variables here?  I'd rather not mention it until we have concrete plans.

bobbrow (approved):

docs/makefile-settings.md
andreeis: Ok I will remove that line and merge the PR.

Issue #142, opened 2021-03-29
Argument parsing is incorrect (e.g. -g file.cpp is grouped as an argument)


sean-mcmanus:
Originally reported at https://github.com/microsoft/vscode-cpptools/issues/7252 .

The argument parsing code doesn't handle the cases where a flag taking a single argument (e.g. -g) is followed by the filename. It's sent to the cpptools extension as "-g file.cpp" which we don't currently support.

Colengms:
Cpptools expect compilerArgs to be individual command line args (as see by the shell).  If an arg is multi-part, each part should be separate.  Note that some args are 3 or 4 part (some Darwin options, such as -sectobjectsymbols, -segaddr, -sectalign, etc.), and secondary args can start with a "-", such as when using -XClang or -Xlinker.

andreeis:
Thank you both for investigating up to the root cause and for giving this extra info, I wasn't aware of this requirement.

MiguelGuthridge:
Sorry, can someone clarify how to fix this? I'm running Ubuntu 20.04 in WSL 2. Building the project works perfectly, both for the extension and for calling `make ...` manually.

I have a project in the default state provided by the extension, and am getting the same errors as above (as well as some with `stdbool.h`). 

I can't find references to `compilerArgs` in any configuration files that I am aware of (`c_cpp_properties.json`, `settings.json`), and I'm not sure where to look for them. Is this be `CFLAGS` in my Makefile? If so, how can I get them to work properly?

Also, I'm unsure why I would be encountering issues like this with the default settings of a reasonably standard project with a reasonably standard Makefile.

Am I doing something wrong here, if so what, and how can I fix it?

andreeis:
@MiguelGuthridge, this is a bug in the Makefile Tools extension and I can't think of a workaround at the moment. We will fix this asap and it will be part of our upcoming 0.2.1 release. Once the fix is in the GitHub repository you will be able to get unblocked before the official release date by manually installing a vsix with the fix. We'll let you know once we fixed this problem.

sean-mcmanus:
A workaround may be to move the "file.cpp" to the end of all your compiler arguments in your makefile.

andreeis:
[...]

PR #143, opened 2021-04-06
compilerArgs entries fix

andreeis:
Fix for bug https://github.com/microsoft/vscode-makefile-tools/issues/142. 
Send different compilerArgs entries for bundled switches or pairs of switch-value if separated by space as a shell would see them.


src/parser.ts
```diff
@@ -414,30 +425,74 @@ function parseAnySwitchFromToolArguments(args: string, excludeArgs: string[]): s
...
+    if (!util.checkFileExistsSync(parseCompilerArgsScriptFile)) {
+        util.writeFile(parseCompilerArgsScriptFile, parseCompilerArgsScriptContent);
+    }
+
+    let scriptArgs: string[] = [];
+    let runCommand: string;
+    if (process.platform === 'win32') {
+        runCommand = "cmd";
+        scriptArgs.push("/c");
+        scriptArgs.push(parseCompilerArgsScriptFile);
```
bobbrow, line 465: Should we surround the file name with double quotes or do you guarantee no spaces in the file name?

bobbrow: similar question for the *nix invocation - might need single quotes? (I forget the rules there)

bobbrow: ok, so the way this is called, you might not need it on windows.  But I still have a concern about linux.

src/parser.ts
```diff
@@ -414,30 +425,74 @@ function parseAnySwitchFromToolArguments(args: string, excludeArgs: string[]): s
...
-        }
+        };
 
-        match1 = match2;
+        let stderr: any = (result: string): void => {
+            logger.messageNoCR(`Error while running the compiler args parser script '${parseCompilerArgsScriptFile}'` +
+                               `for regions "${compilerArgRegions}": "${result}"`, "Normal");
+        };
+
+        const result: util.SpawnProcessResult = await util.spawnChildProcess(runCommand, scriptArgs, vscode.workspace.rootPath || "", stdout, stderr);
```
bobbrow, line 491: `vscode.workspace.rootPath` is deprecated and should not be used.  We'll need to get this value out of `vscode.workspace.workspaceFolders`

src/parser.ts
andreeis: Yes a file with spaces was not able to work. I fixed on linux with single quotes around file name but on windows it's broken too, but adding double quotes around file name is not helping. Investigating how to fix on windows too.
Luckily, the extension is not generating name with space but if TEMP happens to be defined to a path with space then we would have this problem. Thanks for thinking of this scenario.

andreeis: I fixed windows as well (I was sending parameters of script as parameters of cmd.exe and confused it).
I tested these diffs again on both.

src/parser.ts
andreeis: Right, done.

bobbrow (approved):

Issue #134, opened 2021-02-24
Do not append .exe to command


Marsfan:
I primarily use makefiles though WSL. Because of this, the way I run `make` is with the following command

```powershell
wsl -e make
```

When I configure the makePath as 
```JSON
 `"makefile.makePath": "wsl -e make",`
```
The extension then adds the the extension `.exe` to the end of the command. This results in the following error in the _Makefile Tools_ output window:

```powershell
'"wsl -e make.exe"' is not recognized as an internal or external command,
```

andreeis:
Thank you for opening this report. We will fix this for the soon upcoming patch release.
The problem is that we thought of makePath as path, not as a command but we understand your scenario and we'll think what to do about it.

As a workaround, make yourself a full clean build log by running any commands you need in your terminal, point that output to the extension via the setting "makefile.buildLog". This will avoid invoking "make" for setting up IntelliSense in your code base.

Then, the targets detection will fail because even if "makefile.buildLog" is used, we still invoke a particular "make" command to produce a special ouput from which we parse targets. But "all" is always included. If you need any other target than "all", define a configuration with "makeArgs" in "makefile.configurations" and pass the desired target as an argument to make.

And then of course the build feature will not work but you can run that from the terminal.
Debug/Run should work fine despite this.

We hope you can still use this extension until we release a fix.

bobbrow:
Have you tried using the [Remote - WSL extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) in VS Code to do your WSL work?  This will launch the extension in linux mode which wouldn't require prepending `wsl -e` to your make command.

Since the release of that extension, the C/C++ extension no longer supports IntelliSense for WSL when running the Windows version of the extension.  (it may still work, but there are no guarantees)

andreeis:
Implementing the possibility to prefix the make invocation command will be addressed later (https://github.com/microsoft/vscode-makefile-tools/issues/152). The 0.2.1 for this GitHub issue will be just being more clever when adding the ".exe" extension on windows. There was some wrong logic around that. PR coming soon.
This fix may also enable you to use a wrapper script (define "makefile.makePath": "./wrapperMake.bat") (which is still a hack but better maybe) which can run a command like "wsl -e make %*" (%* is to pass any arguments given by the extension to "makefile.makePath" forward towards the real make.

andreeis:
[...]

Issue #114, opened 2021-02-18
makefile.extensionOutputFolder should not log user paths to telemetry


bobbrow:
Before I can enable this telemetry point in our dashboard we need to scrub it like some of the other fields that have PII.  It is currently blocked.  We don't need that data anyway, we just care about whether the setting is used or not.

andreeis:
Also fix the debug/launch string in the configure trigger because it's seen as a path.

andreeis:
When I change the default for extensionOutputFolder I get "...". I can't make it report a path. I wonder how to reproduce this. Do you still see this happening?

PR #153, opened 2021-04-28
More 0.2.1 bug fixes

andreeis:
This addresses https://github.com/microsoft/vscode-makefile-tools/issues/134, https://github.com/microsoft/vscode-makefile-tools/issues/114 and one more file quoting scenario during pre-Configure (there's a wrapper script that we generate that needs to call the final target in quotes and also quote an output file which results from running it).


andreeis:
Closing PR and starting from a fresh new branch for simpler diffs.

Issue #146, opened 2021-04-18
launchConfigurations are sometimes duplicated


hitman-codehq:
Hello plugin developers!

I have attached a .vscode\settings.json file that I use for cross development using the makefile plugin (as a .zip file as .json is not a supported upload format).  This settings.json file is configured to allow me to remotely execute the software I have built.

It works well *but* there is sometimes some strange behaviour of the plugin.  If you select the Makefile plugin icon on the left of VS Code and then click the "Change launch configuration" button next to "Launch target" then you should see something like the following in the dropdown (paths probably altered for your system):

\cygdrive\d\Source\WRC\Debug>WRC.debug()
d:\Source\WRC>..\..\Utils\Windows\RADRunner.exe(vampire,script,..\SendAndExecute.rad)

So first, I have no idea what that first \cygdrive line is because it's not in my configuration and I'm not using Cywin (although the cross compiler I am using is based on Cygwin but I never told the plugin that).  What is that exactly?  Selecting it adds an unusable entry to my settings.json file.

The second line, starting with d:\Source... works for me.  However, let's say I want to edit it.  I want to add a parameter to the binaryArgs array for instance.  If I then select the dropdown again then it will display the two entries shown above *and* a new entry representing the target that contains the new parameter.  The old entry shouldn't be present as it doesn't exist any more.

What is more unusual is that if I select the old entry in the dropdown list, it is *recreated* in my settings.json file!

This might seem like a mild annoyance (and I am guessing you might even reply with "It's designed to work like that") but it becomes much more than a mild annoyance when you are making lots of changes to the settings.json file to try to get it working.  Because every change results in a new entry appearing in the dropdown menu and because they all look very similar (or sometimes even the same) it is hard to know which entry actually represents the most recent edit of the settings.json file.

End the end I end up having to quit VS Code and then restart it so that there is only one entry in the dropdown representing the latest "version" of my launch configuration!

It would be very nice if we could fix this, and possibly address the mystery \cygdrive launch target.

[settings.zip](https://github.com/microsoft/vscode-makefile-tools/files/6330922/settings.zip)

andreeis:
I agree that the launch targets symptom that you described can easily become annoying. We have several other similar bugs identified in this area that we were planning to address after 0.2.1 but let me see if I can easily fix this particular scenario for the upcoming 0.2.1.

[...]

Issue #136, opened 2021-02-27
"Makefile: Build clean the current target" is misleading


hitman-codehq:
Hello make plugin team!

Plugin version tested: 0.2.0

Can I suggest a renaming of the command "Makefile: Build clean the current target" as this is quite a misleading name.  I was searching for how to do a "make clean" and could not find it as I was expecting it to be something like "Makefile: Clean the current target".

There are actually two clean commands:

1) "Makefile: Build clean the current target"
2) "Makefile: Build clean the target ALL"

These behave inconsistently and their naming is a bit misleading.  The first one ONLY cleans the current target but the second one cleans and then *also* builds the ALL target.  Given that their naming is almost identical I would expect them to do almost identical things.  Also, "build clean" is sort of broken English and doesn't really say what it will actually do.  Could I suggest something like:

Makefile: Clean the current target
Makefile: Clean the target ALL

Makefile: Clean and then build the current target
Makefile: Clean and then build the target ALL

These are then consistent and will behave in exactly the way that their titles suggest, and are much easier to find.  Their naming is also consistent with the other two build commands ("Makefile: Build the current target" and "Makefile: Build the target ALL").

Many thanks for the great plugin and I hope my little usability suggestion makes some sense!

andreeis:
Thank you for the naming suggestions. I understand the grammar behind them. We will think about it.
In the meantime, I want to explain more from our side and to double check with you how these commands work.

You say that "Build clean the current target" only does a "clean", while "Build clean the target ALL" does a "clean" and builds "all". Actually, "Build clean the current target" is supposed to do a "clean" and build the current target (same approach as the other command). Let me know if that doesn't happen for you. Regardless of the naming, that would be a bug we'd like to fix asap.
But there's a catch. It also depends on target you have defined currently. If your current target is "MyTarget", then the command "Build clean the current target" would generate a command like "make clean MyTarget". If your current target is set to "all", it would be identical to the other command. But if you didn't set yet a target, the default is considered empty since also "make" invoked with no target is actually able to identify a default. But I realize now that when you clean the default target, in the end you have only clean: "make clean".

I think we can fix this by invoking 2 make commands: "make clean" and "make", to ensure cleaning and building the default target. What do you think of this approach?

hitman-codehq:
Hello Andreea.

I finally did some experimentation and yes, I think that invoking with 2 make commands would be better.

[...]

PR #158, opened 2021-05-04
Launch targets/config bug fixes. Invoke make separately for clean.

andreeis:
Launch targets/config bug fixes: issue https://github.com/microsoft/vscode-makefile-tools/issues/146.
Invoke make separately for clean: issue https://github.com/microsoft/vscode-makefile-tools/issues/136.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -410,7 +410,6 @@ async function readCurrentLaunchConfiguration(): Promise<void> {
     if (currentLaunchConfiguration) {
         launchConfigStr = launchConfigurationToString(currentLaunchConfiguration);
         logger.message(`Reading current launch configuration "${launchConfigStr}" from the workspace state.`);
-        statusBar.setLaunchConfiguration(launchConfigStr);
     } else {
         // A null launch configuration after a non empty launch configuration string name
         // means that the name stored in the project state does not match any of the entries in settings.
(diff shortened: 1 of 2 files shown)


src/configuration.ts
```diff
@@ -397,21 +399,34 @@ function getLaunchConfiguration(name: string): LaunchConfiguration | undefined {
...
+        // Make sure to unset the launch configuration in this scenario.
+        if (currentLaunchConfigurationName !== undefined && currentLaunchConfigurationName !== "") {
+            logger.message(`Launch configuration "${currentLaunchConfigurationName}" is no longer defined in settings "makefile.launchConfigurations".`);
+            await setLaunchConfigurationByName("");
+        } else {
+            logger.message("No current launch configuration is set in the workspace state.");
+        }
     }
+
+    statusBar.setLaunchConfiguration(launchConfigStr);
```
bobbrow, line 428: this looks like it is called already on line 413.  Does it need to be called again here?

src/configuration.ts
```diff
@@ -1353,7 +1368,8 @@ export async function selectLaunchConfiguration(): Promise<void> {
 
     // In the quick pick, include also any makefile.launchConfigurations entries,
     // as long as they exist on disk and without allowing duplicates.
-    let launchTargetsNames: string[] = launchTargets;
+    let launchTargetsNames: string[] = [];
+    launchTargetsNames = launchTargetsNames.concat(launchTargets);
```
bobbrow, line 1372: Easier way to copy an array:

```
let launchTargetsNames: string[] = [...launchTargets];
```

src/configuration.ts
andreeis: Actually it needs to be called here and not on 413, thanks for noticing this.

bobbrow (approved):

Issue #126, opened 2021-02-24
Invoke tools with english language


andreeis:
Until we implement localization, we can try to force english on "make" to ensure we always get english output that we can parse, on all systems.
Investigate whether this is possible.

andreeis:
Experiment from GitHub report https://github.com/microsoft/vscode-makefile-tools/issues/116:
$ make tfred
make: *** Keine Regel, um „tfred“ zu erstellen. Schluss.
$ LANGUAGE=en_US make trffef
make: *** No rule to make target 'trffef'. Stop.

andreeis:
Use https://github.com/microsoft/vscode-cmake-tools/pull/1824 as inspiration.

PR #159, opened 2021-05-04
Force English language when needed

andreeis:
Force English when running executables or scripts that need to parse English words from the execution output.
This fixes https://github.com/microsoft/vscode-makefile-tools/issues/126.


bobbrow (approved):
:shipit:

PR #177, opened 2021-05-25
Dev/andris/makefile tools/pr107 compile commands

andreeis:
Based on https://github.com/microsoft/vscode-makefile-tools/pull/107. 

I only did 2 changes: I thought that the boolean setting exportCompileCommandsFile was redundant. As we also have other similar settings that control generating or not an output file (like extensionLog), it's enough to have a string variable with a ".vscode/compile_commands.json" default. When not defined in settings.json, we don't generate anything. When the user starts to write it in the settings.json, the default pops up.

Other changes were around some refactoring of what arguments we pass from parser.ts helpers into cpptools.ts custom provider construction APIs and optimize some forEach loops.

To view how the differences look like between my suggestion and the original PR, see https://github.com/rapgenic/vscode-makefile-tools/pull/1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -440,14 +440,9 @@
                     "description": "Clear the output channel at the beginning of a build",
                     "scope": "resource"
                 },
-                "makefile.exportCompileCommandsFile": {
-                    "type": "boolean",
-                    "default": false,
-                    "description": "Enable generation and exporting of compile_commands.json"
-                },
                 "makefile.compileCommandsPath": {
                     "type": "string",
-                    "default": "compile_commands.json",
+                    "default": ".vscode/compile_commands.json",
                     "description": "The path to the compilation database file",
                     "scope": "resource"
                 }
--- a/package.json
+++ b/package.json
@@ -442,7 +442,7 @@
                 },
                 "makefile.compileCommandsPath": {
                     "type": "string",
-                    "default": ".vscode/compile_commands.json",
+                    "default": null,
                     "description": "The path to the compilation database file",
                     "scope": "resource"
                 }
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -439,6 +439,12 @@
                     "default": true,
                     "description": "Clear the output channel at the beginning of a build",
                     "scope": "resource"
+                },
+                "makefile.compileCommandsPath": {
+                    "type": "string",
+                    "default": ".vscode/compile_commands.json",
```
bobbrow, line 445: It would seem that we are always writing this file as a result. We may need a separate setting to enable/disable it.  Another option is to default this to empty string and not do the compile commands unless it is explicitly set.  This could also be enforced with the `inspect` API.  The advantage of using that API is that vscode will autofill the default when you override it (I think?)

src/parser.ts
```diff
@@ -781,19 +791,23 @@ export interface CustomConfigProviderItem {
...
+// a compile_commands.json file.
 export async function parseCustomConfigProvider(cancel: vscode.CancellationToken, dryRunOutputStr: string,
                                                 statusCallback: (message: string) => void,
                                                 onFoundCustomConfigProviderItem: (customConfigProviderItem: CustomConfigProviderItem) => void): Promise<number> {
     if (cancel.isCancellationRequested) {
         return make.ConfigureBuildReturnCodeTypes.cancelled;
     }
 
-    logger.message('Parsing dry-run output for CppTools Custom Configuration Provider.', "Normal");
+    logger.message('Parsing dry-run output for CppTools Custom Configuration Provider and compile_commands.json generation.', "Normal");
```
bobbrow, line 810: I don't think we need to add this extra text here.

andreeis: Agree.

package.json
andreeis: My mistake. Setting it to null then, to match the extensionLog setting. No writing it unless the user specifies a path.

PR #187, opened 2021-06-22
Small set of regression tests (win, linux)

andreeis:
Didn't look at porting those on mac yet and the mingw port is ongoing and coming soon.


src/make.ts
```diff
@@ -456,15 +459,16 @@ export async function generateParseContent(progress: vscode.Progress<{}>,
         }
         dryrunFile = util.resolvePathToRoot(dryrunFile);
         logger.message(`Writing the dry-run output: ${dryrunFile}`);
-        util.writeFile(dryrunFile, configuration.getConfigurationMakeCommand() + " " + makeArgs.join(" ") + "\r\n");
+        const lineEnding: string = (process.platform === "win32" && process.env.MSYSTEM === undefined) ? "\r\n" : "\n";
```
bobbrow, line 462: Should we just make life easier on ourselves and only use `\n` line endings?

bobbrow:
You will also want to add the CI Actions to this PR (in the .github folder) so we can make sure that the tests run as part of PR.  I saw specific versions of MSVC tools in the baselines, which concerns me unless we have a way to be agnostic about the version number.

src/test/runTest.ts
```diff
@@ -27,11 +27,9 @@ import * as testRunner from 'vscode-test/out/runTest';
             testRunnerPath: extensionTestsPath,
             testWorkspace: reproRootPath
         };
```
bobbrow, line 29: Need to add:
```
testRunnerEnv: {
                "MAKEFILE_TOOLS_TESTING": "1"
            },
```
to the TestOptions

andreeis:
In the end, I ended up disabling extensions from being installed and I updated the C/C++ standard baseline to c11 which is picked when CppTools is older than v4 or not installed. I started to see again standard diffs under the debugger (not in command line) even if there was no --disable-extensions defined in launch.json. 

Tests are passing now, in VSCode debugger and in command line, with the occasional compilerArgs diffs which we will investigate with work item https://github.com/microsoft/vscode-makefile-tools/issues/191.

andreeis:
Forgot to look at windows after the latest changes. Fixing now.

package.json
```diff
@@ -564,7 +564,7 @@
     },
     "scripts": {
         "vscode:prepublish": "yarn compile",
-        "compile": "tsc -p ./",
+        "compile": "yarn install && tsc -p ./",
         "watch": "tsc -watch -p ./",
         "pretest": "yarn compile",
         "test": "node ./out/src/test/runTest.js"
```
bobbrow, line 570: I still think it would be good to have this do `yarn run pretest && node ...`.  Then update the ci.yml's to just call `yarn run test`

src/test/fakeSuite/Repros/.vscode/preconfigure.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
+set PATH="%cd%\bin\InterestingSmallMakefile\CL\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\12.34.56789\bin\Hostx86\x86";%PATH%
```
bobbrow, line 2: does it have to be such a long path, or can it just be `bin\windows\compilers` or something like that?

src/test/fakeSuite/extension.test.ts
```diff
@@ -1,286 +1,330 @@
...
+            // from this extension repository instead of a real VS installation that happens to be in the path.
+            configuration.setPreConfigureScript(path.join(vscode.workspace.rootPath || "./", ".vscode/preconfigure.bat"));
+            await make.preConfigure(make.TriggeredBy.tests);
+
+            configuration.prepareConfigurationsQuickPick();
+            configuration.setConfigurationByName("InterestingSmallMakefile_windows_configDebug");
+            const retc: number = await make.cleanConfigure(make.TriggeredBy.tests, true);
+
+            configuration.setBuildBeforeLaunch(false);
+            const launchConfigurations: string[] = ["bin\\InterestingSmallMakefile\\ARC H3\\Debug\\main.exe(str3a,str3b,str3c)",
```
bobbrow, line 92: This one is not checked in but the others are. Do we need these exe's checked in, or can we .gitignore them?

src/test/fakeSuite/extension.test.ts
```diff
@@ -1,286 +1,330 @@
...
+            // InterestingSmallMakefile_windows_configRelSpeed constructs a more interesting build command.
+            configuration.setTargetByName("Execute_Arch3");
+            make.prepareBuildTarget("Execute_Arch3");
+
+            // Compare the output log with the baseline
+            // TODO: incorporate relevant diff snippets into the test log.
+            // Until then, print into base and diff files for easier viewing
+            // when the test fails.
+            let parsedPath: path.ParsedPath = path.parse(extensionLogPath);
+            let baselineLogPath: string = path.join(parsedPath.dir, "../InterestingSmallMakefile_windows_baseline.out");
```
bobbrow, line 125: I don't understand the `..` here.  I thought the logs and .outs were in the same folder.

src/test/fakeSuite/extension.test.ts
```diff
@@ -1,286 +1,330 @@
...
+            let extensionLogContent: string = util.readFile(extensionLogPath) || "";
+            let baselineLogContent: string = util.readFile(baselineLogPath) || "";
+            let extensionRootPath: string = path.resolve(__dirname, "../../../../");
+            baselineLogContent = baselineLogContent.replace(/{REPO:VSCODE-MAKEFILE-TOOLS}/mg, extensionRootPath);
+            // fs.writeFileSync(path.join(parsedPath.dir, "base.out"), baselineLogContent);
+            // fs.writeFileSync(path.join(parsedPath.dir, "diff.out"), extensionLogContent);
```
bobbrow, line 131: These lines seem useful. Maybe they can be controlled to run via a command line switch

src/test/fakeSuite/extension.test.ts
```diff
@@ -1,286 +1,330 @@
...
+            expect(extensionLogContent).to.be.equal(baselineLogContent);
+        });
+    }
+
+    // dry-run logs for https://github.com/FidoProject/Fido.git
+    if (systemPlatform === "linux" || systemPlatform === "mingw") {
+        test(`Fido - ${systemPlatform}`, async() => {
+            // Settings reset from the previous test run.
+            extension.getState().reset(false);
```
bobbrow, line 205: A lot of the linux test code looks like same between test cases.  Can you open an issue to look into refactoring some of this into a shared, parameterizable method?

andreeis: Yes, added a few improvement bullets in issue https://github.com/microsoft/vscode-makefile-tools/issues/193. After more triaging and discussion I can split that into more, especially if targeting different milestones.

src/test/fakeSuite/extension.test.ts
andreeis: I realized that the output of the tests are already highlighting some useful diffs. I had those lines written for when I used asserts and when my tests were not completely correctly wired up. Then I was not seeing anything and I used these files a lot. I noticed that I didn't use them since I moved to expect and I wired the tests (there was a missing "main();" in runTests.ts). I'm also thinking not to remove the commented code yet. But not activate it under a switch yet.

src/make.ts
andreeis: Yes I ended up doing that. Plus some extra \r\n ==> \n replacements in baseline and runtime logs, because still some were created by various string operations or... I didn't find exactly what. The replacements are also helping in future for when I update the baselines, so that I copy from whenever without impacting the comparison.

package.json
andreeis: Done.

src/test/fakeSuite/Repros/.vscode/preconfigure.bat
andreeis: I only wanted to keep as close to a real VS compiler path. Also, there are a few characteristics that are needed, where parsing could go wrong. I need the space " ", I need "c" after a "\"...etc.... Let's leave it as is for now.

src/test/fakeSuite/extension.test.ts
andreeis: It's just how I set it up. I defined the extension log to be put in the .vscode of the fake repro since it's something that is always regenerated. It was only habit. If I changed that location then I wouldn't need the "..", indeed.

src/test/fakeSuite/extension.test.ts
andreeis: We need them checked in. Depending on finding them or not, we parse extra launch targets (the ones which are directly invoked, with or without parameters, in the makefile). When these binaries are missing, we still detect one launch target for each, based on what the linker produces. We have information from the -out or -o switches of any linker that we support. But when the makefile invokes one such built binary with particular arguments, then we need to parse that invocation from the dryrun log. And we need to check if it exists on disk to be able to reject other lines where there isn't such built binary invocation but something else which happens to have the name of that binary target in the middle. It's a limitation of the extension not being able to smartly figure out a path versus non path to be able to reject some matches.

src/test/fakeSuite/extension.test.ts
bobbrow: ok

src/test/fakeSuite/extension.test.ts
bobbrow: So are you going to check in `bin\\InterestingSmallMakefile\\ARC H3\\Debug\\main.exe`?

bobbrow (approved):

src/test/fakeSuite/extension.test.ts
andreeis: It is already. You're saying you see it is not?
Since we're talking about these exe's being checked in, maybe I should un-checkin one of them and update baselines. When the binary is not there, we can't parse safely the invocations of that binary with arguments. But for a future version.

bobbrow: It wasn't in this PR.  If it's already checked in, then everything should be fine.

Issue #178, opened 2021-05-31
Expose build command to use in tasks


kmbgit:
If my makefile (written for gmake) runs a compiler (gcc) then I'd expect the PROBLEMS tab to detect compiler warnings and allow me to click on them to go to the line of code to fix them.  If one makefile does a cd and calls another makefile in another directory then I'd expect VS Code to track that through the Entering/Leaving messages output by gmake so it knows what directory relative paths are relative to.

None of this happens, making this extension a bit useless for software development.

bobbrow:
Thank you for trying out the extension.  This feature request is on our backlog.  You can track it with issue #7.
There is also a way to set this up with `tasks.json` so that you don't need to wait for #7.  We need to write up some documentation/examples for that (it is also recently requested #179).

andreeis:
@kmbgit, besides the problems not detecting warnings, can you describe what other issues you see related to "CD" commands and "entering/leaving"? I suspect there are other problems that happen, otherwise just not having warnings in the problems tab I don't think that would make the extension "a bit useless" :). Do you have a repro project for us?
Can you experiment with this setting "makefile.ignoreDirectoryCommands"? Set it to false.

mtnmbr:
> Thank you for trying out the extension. This feature request is on our backlog. You can track it with issue #7.
> There is also a way to set this up with `tasks.json` so that you don't need to wait for #7. We need to write up some documentation/examples for that (it is also recently requested #179).

Hi,
could you please provide a link to the workarround using task.json? I am not able to get this done with a setup like this:

~~~
{
    // See https://go.microsoft.com/fwlink/?LinkId=733558
    // for the documentation about the tasks.json format
    "version": "2.0.0",
    "echoCommand": true,
    "tasks": [
        {
            "label": "Build makfileproject",
            "type": "shell",
            "command": "${command:makefile.launchTargetPath}",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "problemMatcher": "$gcc"
        }
    ]
}
~~~

andreeis:
@mtnmbr, the launchTargetPath command will resolve to the name of a binary that is built by your project and which you selected as a launch target. This command is used when debugging or launching in the terminal.
You need to generate a make command. Currently, the Makefile Tools extension does not expose a command to reveal how make is invoked according to all the current settings. Let me retitle this GitHub report to track this. Once this is fixed, you would be able to write
[...]

Issue #7, opened 2020-06-27
Add support for problem matchers


bobbrow:
Add a property to `makefile.configurations` objects for problem matcher support. (e.g. $gcc, $msbuild)

PR #221, opened 2021-11-03
Initial support for problem matchers

andreeis:
Implement feature https://github.com/microsoft/vscode-makefile-tools/issues/7.

Didn't test yet with relative paths of multiple makefile projects in various folders, as commented in https://github.com/microsoft/vscode-makefile-tools/issues/178. Will see how this works for such complex setup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,6 +196,25 @@
                 "title": "Makefile: Reset the Makefile Tools Extension workspace state (For troubleshooting)"
             }
         ],
+        "problemMatchers": [
+            {
+              "name": "gcc",
+              "source": "gcc",
+              "owner": "makefile-tools",
+              "fileLocation": [
+                "autoDetect",
+                "${workspaceFolder}"
+              ],
+              "pattern": {
+                "regexp": "^(.*?):(\\d+):(\\d*):?\\s+(?:fatal\\s+)?(warning|error):\\s+(.*)$",
+                "file": 1,
+                "line": 2,
+                "column": 3,
+                "severity": 4,
+                "message": 5
+              }
+            }
+          ],
         "configuration": {
             "type": "object",
             "title": "makefile",
@@ -231,6 +250,14 @@
                                     "type": "string"
                                 },
                                 "default": []
+                            },
+                            "problemMatchers": {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "default": [],
+                                "description": "Problem matcher names to use when building the current target"
                             },
                             "buildLog": {
                                 "type": "string",
@@ -336,9 +363,9 @@
                     "scope": "resource"
                 },
                 "makefile.makeDirectory": {
-                  "type": "string",
-                  "description": "The folder path to be passed to make via the switch -C",
-                  "scope": "resource"
+                    "type": "string",
+                    "description": "The folder path to be passed to make via the switch -C",
+                    "scope": "resource"
                 },
                 "makefile.makefilePath": {
                     "type": "string",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -196,6 +196,25 @@
                 "title": "Makefile: Reset the Makefile Tools Extension workspace state (For troubleshooting)"
             }
         ],
+        "problemMatchers": [
+            {
+              "name": "gcc",
```
bobbrow, line 201: Will this clash with the C++ extension's problem matcher?  I thought we would just insert the correct regex before starting the task if someone used `$gcc` as their problem matcher. Otherwise, I would worry that vscode doesn't know what to do with two `$gcc` entries.

package.json
```diff
@@ -232,6 +251,14 @@
                                 },
                                 "default": []
                             },
+                            "problemMatchers": {
```
bobbrow, line 254: I expected that we have the same schema as the vscode problem matchers for tasks here. We want to support more than just named problem matchers. Users should be able to specify their own regex patterns too.

src/make.ts
```diff
@@ -326,14 +328,32 @@ export async function doBuildTarget(progress: vscode.Progress<{}>, target: strin
             logger.messageNoCR(result, "Normal");
         };
 
-        // The build invocation should use the system locale.
-        const result: util.SpawnProcessResult = await util.spawnChildProcess(configuration.getConfigurationMakeCommand(), makeArgs, util.getWorkspaceRoot(), false, stdout, stderr);
-        if (result.returnCode !== ConfigureBuildReturnCodeTypes.success) {
+        let myTaskCommand: string = `${configuration.getConfigurationMakeCommand()} ${makeArgs.join(" ")}`;
```
bobbrow, line 331: Does `makeArgs.join` do the right thing for args with spaces in them or do we need to surround spaced args with quotes before doing this?  Alternately, is there a function on the `shlex` module that can compose the command line for us?  Or does the `vscode.ShellExecution` support passing in the args as an array so we don't have to do it?

andreeis: Great idea, found that ShellExecution can make a clear separation between command (which is make) and args and has settings to properly quote each of them.

package.json
andreeis: Oh, I forgot to reply to this one. According to what I read in VSCode problem matcher documentations (and also in issue comment https://github.com/microsoft/vscode/issues/61497#issuecomment-456320050), the last one that is parsed wins.
On one side, who uses Makefile Tools without CppTools? So maybe we should remove gcc from package.json.
But if that happens, having one gcc here as a backup doesn't hurt. Let's discuss in the morning meeting.

bobbrow: If last one wins and VS Code doesn't log errors/warnings, then we can keep this one.  There are actually quite a few Makefile Tools users who do not use cpptools (14%)

package.json
andreeis: I will start a conversation with the VSCode team (I didn't open a GitHub issue yet about API to create a task with a problem matcher based on it's complete schema versus just a string name, will do today). Since currently it is not supported, even if they approve the request it will be part of the next VSCode release, so in the mean time we will have to accept this limitation.

A global problem matcher defined in .vscode/tasks.json does not have a name or label. But we can do something like simply read it, assign it a name that we use when we create the task. Let's discuss in the morning meeting.

andreeis:
New feature request for the VSCode team opened here: https://github.com/microsoft/vscode/issues/137221.
This will help us implement the possibility for our users to define their own new problem matchers without needing an update of Makefile Tools.

andreeis:
Squash and Merge. Reviewed with @bobbrow.

Issue #119, opened 2021-02-23
Align C/C++ standard defaults to CppTools and CMake Tools


andreeis:
We need to fix the C/C++ standard defaults that the extension sends to CppTools.
This is the same fix as CMake Tools needs to do soon: https://github.com/microsoft/vscode-cmake-tools/issues/1477.

Issue #115, opened 2021-02-19
Output window is opened whenever the extension processes changes to the makefile


dan1994:
First of all, this extension is awesome! I've been waiting for makefile support for some time now!

I've been tinkering with a sample project with a makefile and a single C++ source. What I've noticed is that whenever I update the makefile and then focus on the C++ file, the Makefile Tools output window is opened.
Is there a way to suppress this, as it is pretty annoying, especially when  simultaneously using the terminal (it switches from the terminal to the output window)?

andreeis:
Thank you for opening this report. We tried to tune the frequency with which the the extension shows the output window, not too often to be annoying but also to mark important moments in the project. This is good feedback and we can adjust at any time.
Our thought was that it would be useful to display the output window at the beginning of build, pre-configure and configure operations.
When you modify the makefile, you trigger an automatic re-configure of your project (because maybe the edits will translate into new switches, source files.... that would impact the IntelliSense) and this is a visibility trigger.

First, let us know what you think about this approach. Is this still too often? 

We can always add a setting to disable this, just in case.

What you can do now is to disable indirectly the output channel reveal by disabling automatic configures, with setting makefile.configureOnEdit as false. But remember to invoke the command makefile.configure (from the palette) every time you think you made a makefile edit that should change your IntelliSense state.

Let us know your thoughts about the visibility triggers and otherwise we'll still keep this work item opened and triage it because a new setting will not hurt.

dan1994:
Thanks for the quick reply!

My personal preference is to keep things in the background unless something goes wrong. In the ideal scenario I don't need to care when the extension is triggered nor what it does. All I want to care about is that the include directory or the new define I added are recognized by the intellisense. If something doesn't work I'll just open the output tab manually.

I do think that adding a setting to turn this off is enough, and will definitely resolve the issue as far as I'm concerned.

andreeis:
@dan1994, in the end we didn't add a new setting but we made the output channel to show only when there is an error.
Until we release Makefile Tools 0.2.1, if you want you can install this vsix to benefit of this change: https://github.com/microsoft/vscode-makefile-tools/actions/runs/793552744.

dan1994:
Thank you!

COOLIRON2311:
> @dan1994, in the end we didn't add a new setting but we made the output channel to show only when there is an error.
[...]

PR #227, opened 2021-11-08
Small 0.3 bug fixes (c/c++ standard defaults and output channel)

andreeis:
Fix for [Don't hardcode any c/c++ standard defaults and let CppTools query the compiler.](https://github.com/microsoft/vscode-makefile-tools/issues/119) and [Don't show output channel when project is loaded in VSCode](https://github.com/microsoft/vscode-makefile-tools/issues/115).

Also don't show the Output Channel when VSCode opens a c/c++ source file for which Makefile Tools did not find an IntelliSense configuration for.


src/parser.ts
```diff
@@ -1455,12 +1455,9 @@ function parseStandard(cppVersion: cpp.Version | undefined, std: string | undefi
...
     if (!std) {
-        // Standard defaults when no std switch is given
-        if (language === "c") {
-            return "c11";
-        } else if (language === "cpp") {
-            return "c++17";
-        }
+        // Return an undefined standard default when no std switch is given
+        // so that CppTools query the compiler itself for these defaults.
+        return undefined;
```
bobbrow, line 1460: Double check the cpp.Version if this can be done.  It might need to be v5.

andreeis: Thanks, great catch. While following up on that I noticed other things that I missed when bringing CMake Tools equivalent over.

bobbrow (approved):

PR #231, opened 2021-11-15
Version update

andreeis:
My latest attempt to fix the VSIX build custom action (which was done directly in main by mistake) reported some error about 0.2.2 version so I update it now to 0.3 since we have to do that anyway soon.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,14 +26,6 @@ jobs:
           node-version: '12'
           check-latest: true
   
-      - name: Update the package version
-        uses: jacobtomlinson/gha-find-replace@master
-        with:
-          find: "\"version\": \"(\\d+\\.\\d+\\.\\d+)\""
-          replace: "\"version\": \"$1-${{ steps.get-short-sha.outputs.sha }}\""
-          include: "package.json"
-          exclude: ".git"
-
     # Runs a single command using the runners shell
       - name: Build the VSIX
         run: |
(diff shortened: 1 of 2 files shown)


andreeis:
Works finally.

.github/workflows/main.yml
```diff
@@ -26,14 +26,6 @@ jobs:
           node-version: '12'
           check-latest: true
   
-      - name: Update the package version
```
bobbrow, line 29: We need to keep this or else you'll have to change the version in package.json for every VSIX you build.  I don't understand how this could be causing the TypeScript compilation error. It's not even in the same step.

bobbrow: The problem is that the previous task was missing.  This is what needs to be here:
```
      - name: Get Short SHA
        id: get-short-sha
        run: echo "::set-output name=sha::$(git rev-parse --short ${GITHUB_SHA})"

      - name: Update the package version
        uses: jacobtomlinson/gha-find-replace@master
        with:
          find: "\"version\": \"(\\d+\\.\\d+\\.\\d+)\""
          replace: "\"version\": \"$1-${{ steps.get-short-sha.outputs.sha }}\""
          include: "package.json"
          exclude: ".git"
```

bobbrow: Oh, wait, we never had these two steps in the build process here.  It must have been added partially outside of PR then.  Sorry, I got confused.  But if you want to add the short SHA to the version like we do for CMake Tools, copy those two actions in there.

Issue #219, opened 2021-10-05
command-line error: language modes specified are incompatibleC/C++(1027)


lab313ru:
My project has the following settings:
```
"cStandard": "c99",
"cppStandard": "c++03",
"intelliSenseMode": "windows-gcc-x86",
"configurationProvider": "ms-vscode.makefile-tools"
```

After every reopening of the project this problem happens and doesn't allow to build anything. The only solution is to disable then enable the extension. After that everything does work. But after you'll close VScode the problem will appear again.

andreeis:
@lab313ru, thank you for creating this issue.
Can you set "makefile.loggingLevel" to "Debug" and "makefile.extensionLog" to "extension.log" in your project .vscode/settings.json, reload your project in VSCode, let it finish configuring (you should see something like "configure finished" in the Makefile Tools output channel), then share with us the files .vscode/dryrun.log and .vscode/extension.log?
And let us know what source file exhibits this problem, so we track that file name in the logs.

Also, does the same happen if you open the c/c++ source file after configure finishes? I am asking in case there may be a bug about updating the IntelliSense information after configure finishes, when a file is opened before we have that available.

If your project is easy to share with us entirely, we can do all the investigation on our side.

lab313ru:
[dryrun.log](https://github.com/microsoft/vscode-makefile-tools/files/7289716/dryrun.log)
[extension.log](https://github.com/microsoft/vscode-makefile-tools/files/7289717/extension.log)

It's happens for `core.c` file at its first line.

What I'm doing to reproduce is:
1. Open VS Code (by default it opens the last opened, so the current, project)
2. Problem appears
3. Disable extension, then reload VS Code
4. Enable extension
5. Problem disappears
6. Close then open VS Code to make this problem appear again

andreeis:
@lab313ru, thank you for the files. The disable/enable is a mystery for me (the squiggle should not disappear) but at least I spotted something in the logs you sent.

In what kind of environment you are opening VSCode? Is it MinGW? If yes, Is "cygpath" command working properly in that environment? What does it output for "cygpath \d\SGDK_projects\clock_tower\src\core.c -w" ?

lab313ru:
No MinGW or Cygwin paths were added to PATH. I'm just using a folder with 
all executables I need and passing its path to the variable in my Makefile.

For more information I recommend you to look at the project which 
executables I use: https://github.com/Stephane-D/SGDK/blob/master/makefile.gen

[...]

Issue #194, opened 2021-06-30
"..." prevents proper parsing of source files


andreeis:
@andreeis I have also been having problems with IntelliSense not working on my project.  Luckily I stumbled upon this issue here and this helped me to figure out that it is my makefile that was the problem.  I have this rule:

```
$(OBJ)/%.o: %.c
	@echo Compiling $<...
	$(CC) $(CFLAGS) $(IFLAGS) -o $(OBJ)/$*.o $<

```

It turns out to be those ... characters in the echo line that cause the problem as they prevent the .c file from being found.  The output in dryrun.log is:

echo Compiling MyFile.c...

When I add a space before the ... then it works:

echo Compiling MyFile.c ...

I didn't see any mention in the documentation (https://marketplace.visualstudio.com/items?itemName=ms-vscode.makefile-tools) of the filenames needing to be printed somewhere in the dryrun.log file (without trailing ...!) in order for IntelliSense to work.  IntelliSense configuration with the makefile plugin is rather black magic at the moment, so could I suggest a bit more explanation of how it works in the documentation?  It would help people to debug their problems.

Even better, when outputting the dry run information in the output window, if no files are found, output a message at the end telling the user this, and giving them tip that the names of the files should be listed in the dryrun.log output.  I see that there is positive confirmation when a file is found (Sending configuration for file \cygdrive\d\Source\ClassAct\Examples\Label\LabelExample.c -----------------------------------) but it would be good to also have negative confirmation.

Handling those trailing ... would be nice too, if possible.  That had me scratching my head for a while.  :-)

_Originally posted by @hitman-codehq in https://github.com/microsoft/vscode-makefile-tools/issues/173#issuecomment-871117837_

hitman-codehq:
Hi @andreeis I have been away on holiday but now I am back it is time to continue with this issue.  :-)

I am not really sure how to approach this problem as it is quite complicated.  It seems that on MacOS, everything is working almost perfectly, even with the "..." problem mentioned above.  It is only on Windows that it seems to not work.  Unless the problem has been fixed since I reported it?

The only thing that doesn't work on MacOS is that whenever I follow a symbol I get the following warning:

Configuration for file /Users/colinward/Source/StdFuncs/StdWindow.cpp was not found. CppTools will set a default configuration.

And yet everything is working, including following symbols with F12.  It works whether following a local symbol or whether following a symbol into my cross compiler's includes in /opt/amiga.

Do you have any idea what this message means?

[...]

Issue #156, opened 2021-04-23
Suddenly happened - "command-line error: language modes specified are incompatible"  on all my c code


drimmeer:
Bug type: Language Service

**Describe the bug**
- OS and Version: Windows Server 2012 R2 standard
- VS Code Version: 1.55.2
- C/C++ Extension Version: 1.3.1
- Other extensions you installed (and if the issue persists after disabling them): Make File Tools 0.2.0
- If using SSH remote, specify OS of remote machine: N/A
- A clear and concise description of what the bug is, including information about the workspace (i.e. is the workspace a single project or multiple projects, size of the project, etc).

I have only one very small c project loaded into VScode. It worked well at the beginning.
 But from some time on, it could not go to the 'Reference' or 'Definition' or 'Declaration' of any method. 
Then later, some more serious problem happened: I found in the 'Problem' view that any opened c code was reported there showing error message "command-line error: language modes specified are incompatible"

BTW, the 'Build' runs well successfully. That means the c codes have no problem.

**Steps to reproduce**

1. Click on one c program and open it in the Editor.
2. Go to the 'Problem' view, you will see that it shows error "command-line error: language modes specified are incompatible" which directs you to the first line of the code.

**Expected behavior**

I would like to be able to go to a method referencing places, definition and declaration places. Also there should not be such errors "command-line error: language modes specified are incompatible".

**Code sample and logs**

- Configurations in `c_cpp_properties.json`
   There is no such file.

- content in 'settings.json'
```
{
    "C_Cpp.default.configurationProvider": "ms-vscode.makefile-tools"
}
```

- content in 'configurationCache.log'
```
[...]

PR #234, opened 2021-11-17
Paths processing for non win tools run on win

andreeis:
Possibly addresses https://github.com/microsoft/vscode-makefile-tools/issues/194 (user confirms partial improvement), https://github.com/microsoft/vscode-makefile-tools/issues/156 (user still sees the issue) and https://github.com/microsoft/vscode-makefile-tools/issues/219 (user confirms the problem is fixed).


bobbrow (approved):
One optional comment on the condition to enter the `ensureWindowsPath` function (move the leading `/` check into the function), but otherwise, good to check in if you want to pass on that.

Issue #246, opened 2021-12-06
Makefile Tolls Icon on side bar gone with 0.3.0


baumra:
First of all, thank you for the great extension!

For me with the update to 0.3.0 the Makefile Tools icon on the side bar is gone. When I go back to 0.2.2 it is there again.
Is this intentional?

andreeis:
We had complaints that projects that were not Makefile based would still cause that UI and palette commands to be visible so we made sure that these are now visible only when the extension activates.
This indeed breaks the projects that don't have a makefile in the root, but use settings like "makefile.makefilePath" and "makefile.makeDirectory".
We have work item https://github.com/microsoft/vscode-makefile-tools/issues/5 which addresses more complex activation scenarios that we plan to improve.

In the mean time, what you can do, is simply run the configure command after you open your project in VSCode. Then all will show up. If not, let us know, that would be a bug we should fix asap.

baumra:
Thanks, that explains it. We use "makefile.makefilePath". So thank you for the explanation.
As all of out projects have their Makefile buried somewhere deep in sub-directories, a common solution would be appreciated.
Meanwhile I'll use the workaround running configure manually.

andreeis:
We just released a fix for this problem with Makefile Tools 0.3.1 today. Let us know if you encounter any other issues.

baumra:
Unfortunately this fixed it only partly. The icon is now shown but the build does not work.
I configured some makeArgs in the makefile.configurations like this:
```
"makefile.configurations": [
        {
            "name": "release",
            "makeArgs": ["TARGET=release -sj1", "-C ./src/mycode"]
        }
]
```

The Makefile expects $(TARGET) to contain "release" in this case. But now $(TARGET) contains "release -sj1-C ./src/mycode".
The build subsequently fails.

andreeis:
@baumra, yes let's create a new issue for this.
My first observation is that you should define 4 makeArgs instead of 2. -sj1 should be the second, the folder after -C should be the fourth... Try that and let me know how it works.

baumra:
Ok, a new issue is not needed then. I separated the arguments and it works now. Thank you!

Issue #245, opened 2021-12-03
why do you add build button to any where?


heartacker:
![image](https://user-images.githubusercontent.com/1876302/144565184-5f219d85-173d-42fe-9a28-82dac8474f03.png)

![image](https://user-images.githubusercontent.com/1876302/144565233-ac9acfcc-fc4a-408e-85a1-a8fe34652114.png)

![image](https://user-images.githubusercontent.com/1876302/144565293-1c86924c-f3e7-4258-9b5c-b17e4e85de26.png)

![image](https://user-images.githubusercontent.com/1876302/144565318-aec3bb32-c205-4688-8759-a116a0fdc214.png)

版本: 1.63.0-insider (system setup)
提交: cca3e817a2608225d4ae8715308195c91f765bad
日期: 2021-12-03T05:15:56.178Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.14393
![image](https://user-images.githubusercontent.com/1876302/144565441-82055ce3-ef1e-4c0b-b46e-bc818b69f8d9.png)

addisonElliott:
If I'm understanding OP correctly, he's questioning why the build button is present in all sorts of places in VS Code. I think it's a valid point.

Personally, I don't think the build button should be anywhere besides the Makefile tab.

andreeis:
@heartacker and @addisonElliott, that was a git merge mistake on our part that happened sometime during our latest PRs and we didn't notice. Thank you for catching this. The button will be fixed in the upcoming release, 0.4.

andreeis:
Closing this since we just released a fix with Makefile Tools 0.3.1 today.

Issue #233, opened 2021-11-17
"terminal.integrated.automationShell" is not used when running Configure


daniel-a-melo:
I have a `Makefile` that can only execute in UNIX-like shells (Linux, MSYS2, Cygwin, MacOS). It uses the find function.

```Makefile
TEST_SRC=$(shell find tests -name "*_test.c")
```

When I run configure in Windows using MSYS2, the extension seems to be executing it in a shell where find does not work as expected (is it executing on CMD, PowerShell?)

dryrun.log
```
C:\Users\user\msys64\mingw64\bin\mingw32-make.exe --dry-run --keep-going --print-directory
mingw32-make: Entering directory 'c:/Users/user/c-project'
 

File not found - *_test.c
 
File not found - *_test.c
 
File not found - *_test.c
 
mingw32-make: Leaving directory 'c:/Users/user/c-project'
```

I would expect it to respect `terminal.integrated.automationShell` setting, and run Configure in the configured shell for automation tasks, which is MSYS2's bash. 

vscode settings.json
```json
{
    "terminal.integrated.profiles.windows": {
        "PowerShell": {
            "source": "PowerShell",
            "icon": "terminal-powershell"
        },
        "Command Prompt": {
            "path": [
                "${env:windir}\\Sysnative\\cmd.exe",
                "${env:windir}\\System32\\cmd.exe"
            ],
            "args": [],
            "icon": "terminal-cmd"
        },
        "Bash": {
            "env": {
                "CHERE_INVOKING": "1",
                "MSYSTEM": "MINGW64",
                "MSYS2_PATH_TYPE": "inherit",
                "PATH": "%MSYS2_HOME%\\mingw64\\bin;%PATH%"
            },
            "path": [
                "${env:MSYS2_HOME}\\usr\\bin\\bash.exe"
            ],
            "args": [
                "--login"
            ]
        },
    },
    "terminal.integrated.defaultProfile.windows": "Bash",
    "terminal.integrated.automationShell.windows": "${env:MSYS2_HOME}\\usr\\bin\\bash.exe",
    "makefile.makePath": "C:/Users/user/msys64/mingw64/bin/mingw32-make.exe"
}
```

jankap:
I'm facing exactly the same issue. `find.exe` seems to have precedence of unix `find` which fails my Makefiles.

Same is true when calling `make` directly from `tasks.json` btw.

andreeis:
[...]

PR #247, opened 2021-12-07
Fix for GitHub issues for 0.3.1 release

andreeis:
- Honor the "terminal.integrated.automationShell" setting when spawning make for configure. [#233](https://github.com/microsoft/vscode-makefile-tools/issues/233)
- Remove the "build" button icon from other UIs than the main Makefile Tools panel. [#245](https://github.com/microsoft/vscode-makefile-tools/issues/245)
- The build task fails for projects using -f or -C (makefile not in root) because of quoting. [#249](https://github.com/microsoft/vscode-makefile-tools/issues/249)
- Fix activation for makefiles below the root. [#248](https://github.com/microsoft/vscode-makefile-tools/issues/248)

Duplicates of 248: https://github.com/microsoft/vscode-makefile-tools/issues/250 and https://github.com/microsoft/vscode-makefile-tools/issues/246

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,9 +67,9 @@
         "onCommand:makefile.outline.cleanConfigure",
         "onCommand:makefile.outline.preConfigure",
         "onCommand:makefile.resetState",
-        "workspaceContains:makefile",
-        "workspaceContains:Makefile",
-        "workspaceContains:GNUmakefile"
+        "workspaceContains:**/makefile",
+        "workspaceContains:**/Makefile",
+        "workspaceContains:**/GNUmakefile"
     ],
     "main": "./out/src/extension.js",
     "contributes": {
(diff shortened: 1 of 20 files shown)


bobbrow (approved):

package.json
```diff
@@ -67,9 +67,9 @@
         "onCommand:makefile.outline.cleanConfigure",
         "onCommand:makefile.outline.preConfigure",
         "onCommand:makefile.resetState",
-        "workspaceContains:makefile",
-        "workspaceContains:Makefile",
-        "workspaceContains:GNUmakefile"
+        "workspaceContains:**/makefile",
```
bobbrow, line 70: Somehow I missed this. We weren't supposed to activate the extension for makefiles anywhere in the workspace until we could ensure it wouldn't interfere with CMake Tools. VS Code fixed the activation on timeout problem, but do we do the right thing when CMake generates "Unix Makefiles"?

PR #258, opened 2021-12-16
Initial localization support

andreeis:
Inspired as much as possible from CMake Tools.
A few instances of strings are localized here and there as a proof of concept. I will do a more thorough pass once all this is working.
I know some of the files that are added here will eventually be automatically generated, but I wanted to have as much as possible already in place to see how the scripts are working. I wrote manually all the languages jsons to have a more complete start.

I added a new pipeline following CMake Tools OneLocBuild, but it fails and I don't know yet how to fix. Investigating.
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5556033&view=results

yarn compile-production works fine on my computer, I don't know why the pipeline complains about a missing file. It is checked in.


andreeis:
Converting to draft because I forgot some things I wanted to cleanup before review.

.gitignore
```diff
@@ -1,5 +1,12 @@
...
+node_modules
+.vscode-test/
+*.log
+*.vsix
+vscode-extensions-localization-export/
+vscode-translations-import/
+jobs/loc/LCL
+OneLocBuild
+**/nls.*.json
+**/*.nls.json
```
Colengms, line 11: I think the right pattern to ignore would be `**/package.nls.*.json`

gulpfile.js
```diff
@@ -0,0 +1,306 @@
...
+const sourcemaps = require('gulp-sourcemaps');
+const ts = require('gulp-typescript');
+const typescript = require('typescript');
+const tsProject = ts.createProject('./tsconfig.json', { typescript });
+const filter = require('gulp-filter');
+const vinyl = require('vinyl');
+const jsonc = require('jsonc-parser');
+
+
+// Patterns to find schema files
```
Colengms, line 24: If you don't have any JSON schema files, you could remove the code in here that translates those.

i18n/chs/package.i18n.json
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
```
Colengms, line 1: I believe if the i18n files are not present, vscode-nls will use English strings.  You could omit them for now.  The OneLocBuild pipeline should result in a PR being generated that adds them.

jobs/loc/LocProject.json
```diff
@@ -0,0 +1,14 @@
+{
+    "Projects": [
```
Colengms, line 2: In case you haven't already; You'll need to send an email to loc team to get them to add Loc a project for vscode-makefile-tools.  (I can send you their contact info privately).

jobs/loc/TranslationsImportExport.yml
```diff
@@ -0,0 +1,56 @@
...
+    SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+  inputs:
+    locProj: 'jobs/loc/LocProject.json'
+    outDir: '$(Build.ArtifactStagingDirectory)'
+    isCreatePrSelected: false
+    prSourceBranchPrefix: 'locfiles'
+    packageSourceAuth: 'patAuth'
+    patVariable: '$(OneLocBuildPat)'
+    LclSource: lclFilesfromPackage
+    LclPackageId: 'LCL-JUNO-PROD-VMAKEFILE'
```
Colengms, line 45: Did you get this ID from the loc team?  If not, you'll want to get an ID from them.

jobs/loc/TranslationsImportExport.yml
```diff
@@ -0,0 +1,56 @@
...
+    lsBuildXLocPackageVersion: '7.0.30510'
+
+- task: CmdLine@2
+  inputs:
+    script: 'node ./translations_auto_pr.js microsoft vscode-makefile-tools csigs $(csigsPat) csigs csigs@users.noreply.github.com "$(Build.ArtifactStagingDirectory)/loc" vscode-extensions-localization-export/vscode-extensions'
```
Colengms, line 50: I don't recall, but you might need to provide the csigs account with permission to push a branch and post a PR.

src/configuration.ts
```diff
@@ -582,9 +586,10 @@ export function getCommandForConfiguration(configuration: string | undefined): v
 
         // If configuration command has a path (absolute or relative), check if it exists on disk and error if not.
         // If no path is given to the make tool, search all paths in the environment and error if make is not on the path.
+        const notFoundStr: string = localize("not.found", "not found.");
         if (configurationCommandPath  !== "") {
             if (!util.checkFileExistsSync(configurationMakeCommand)) {
-                vscode.window.showErrorMessage("Make not found.");
+                vscode.window.showErrorMessage(`Make ${notFoundStr}.`);
```
Colengms, line 592: I would use `{0} not found.`.  The more contents you can give the loc team, the better.

src/extension.ts
```diff
@@ -368,7 +372,7 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
 }
 
 export async function deactivate(): Promise<void> {
-    vscode.window.showInformationMessage('The extension "vscode-makefile-tools" is de-activated');
+    vscode.window.showInformationMessage(localize("extension.deactivated", "The extension 'vscode-makefile-tools' is de-activated."));
```
Colengms, line 375: You may want to move the extension name (anything you don't want to see translated) into a variable that gets provided via a `{0}` placeholer.

translations_auto_pr.js
```diff
@@ -0,0 +1,215 @@
...
+let authToken = process.argv[5];
+let userFullName = process.argv[6];
+let userEmail = process.argv[7];
+let locRootPath = process.argv[8];
+let locSubPath = process.argv[9];
+
+if (!repoOwner || !repoName || !authUser || !authToken || !userFullName || !userEmail || !locRootPath || !locSubPath) {
+    console.error(`ERROR: Usage: ${path.parse(process.argv[0]).base} ${path.parse(process.argv[1]).base} repo_owner repo_name auth_token user_full_name user_email loc_root_path loc_sub_path`);
+    console.error(`   repo_owner - The owner of the repo on GitHub.  i.e. microsoft`);
+    console.error(`   repo_name - The name of the repo on GitHub.  i.e. vscode-cpptools`);
```
Colengms, line 26: Maybe update to refer to `vscode-makefile-tools`?

Colengms:
I don't think you want to be running the OneLocBuild pipeline as part of PR validation.  That should be run once daily, and only against the main branch (where the strings are stable).

gulpfile.js
andreeis: We will have in future. We'll spend less time removing now and adding back then. :).

jobs/loc/TranslationsImportExport.yml
andreeis: I just updated after copy paste. Thank you for explaining what needs to happen and I'll contact the loc team once I get their contact from you.

translations_auto_pr.js
andreeis: Ah how I missed that! Of course, all scripts I copied from cmake or cpp tools I updated wherever it was obvious. But I missed this one, great catch, thank you.

andreeis:
> I don't think you want to be running the OneLocBuild pipeline as part of PR validation. That should be run once daily, and only against the main branch (where the strings are stable).

@Colengms, actually I don't remember doing anything specific to add the localization pipeline as part of the PR validation. With every update I push, that check is not the fourth one in the list, I see only the other intended three. But as soon as I go to the pipeline link and run it myself just to see what happens, soon after it starts running it gets automatically added in here. I'll do the same for cmake tools when I have the chance to see that if this behavior is the same or maybe I did link the two items too much together.

src/launch.ts
```diff
@@ -1,360 +1,362 @@
...
+                }
+             }
+         }
+
+        let currentLaunchConfiguration: configuration.LaunchConfiguration | undefined = configuration.getCurrentLaunchConfiguration();
+        if (!currentLaunchConfiguration) {
+            // If no launch configuration is set, give the user a chance to select one now from the quick pick
+            // (unless we know it's going to be empty).
+            if (configuration.getLaunchTargets().length === 0) {
+                vscode.window.showErrorMessage(localize("cannot.op.no.launch.config.list.launch.targets.empty", "Cannot '{0}' because there is no launch configuration set and the list of launch targets is empty. " +
```
bobbrow, line 242: I don't see this key in the generated files. I was worried because of the `+` between strings.  I don't know if that works or if you need to regenerate.

src/make.ts
```diff
@@ -115,7 +119,8 @@ export function blockedByOp(op: Operations, showPopup: boolean = true): Operatio
     if (getIsConfiguring()) {
         // A configure in the background shouldn't block anything except another configure
         if (getConfigureIsInBackground() && op !== Operations.configure) {
-            vscode.window.showInformationMessage(`The project is configuring in the background and ${op} may run on out-of-date input.`);
+            vscode.window.showInformationMessage(localize("project.configuring.background.op.may.run.on.out.of.date.input",
```
bobbrow, line 122: >project.configuring.background.op.may.run.on.out.of.date.input [](http://example.com/codeflow?start=59&length=62)

The key names don't need to summarize the message. I would recommend making them shorter.

src/launch.ts
```diff
@@ -1,360 +1,362 @@
...
+            return LaunchStatuses.blocked;
+        }
+
+        if (configuration.getBuildBeforeLaunch()) {
+            let currentBuildTarget: string = configuration.getCurrentTarget() || "";
+            logger.message(`Building current target before launch: "${currentBuildTarget}"`);
+            let buildSuccess: boolean = (await make.buildTarget(make.TriggeredBy.buildTarget, currentBuildTarget, false)) === make.ConfigureBuildReturnCodeTypes.success;
+            if (!buildSuccess) {
+                logger.message(`Building target "${currentBuildTarget}" failed.`);
+                let noButton: string = localize("No", "No");
```
bobbrow, line 218: all the other key names start with a lower-case letter. I recommend that for Yes/No as well.

src/launch.ts
andreeis: I removed the "+" pattern, just in case.

package.nls.json
```diff
@@ -0,0 +1,19 @@
...
+   "makefile-tools.command.makefile.setBuildTarget.title": "Makefile: Set the target to be built by make",
+   "makefile-tools.command.makefile.setLaunchConfiguration.title": "Makefile: Set the make launch configuration",
+   "makefile-tools.command.makefile.configure.title": "Makefile: Configure",
+   "makefile-tools.command.makefile.cleanConfigure.title": "Makefile: Clean Configure",
+   "makefile-tools.command.makefile.preConfigure.title": "Makefile: Pre-Configure",
+   "makefile-tools.command.makefile.resetState.title": "Makefile: Reset the Makefile Tools Extension workspace state (For troubleshooting)",
+   "makefile-tools.configuration.views.makefile.outline.description": "Project Outline",
+   "makefile-tools.configuration.makefile.makePath.description": "The path to the make tool",
+   "makefile-tools.configuration.makefile.configurations.description": "The user defined makefile configurations",
+   "description": "%makefile-tools.configuration.makefile.m"
```
bobbrow, line 18: something looks wrong with this line

i18n/chs/src/launch.i18n.json
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+// Do not edit this file. It is machine generated.
+{
+	"yes": "Yes",
+	"no": "No",
+   "build.failed.continue.anyway": "Build failed. Do you want to continue anyway?",
+   "cannot.op.no.launch.config.targets": "Cannot '{0}' because there is no launch configuration set and the list of launch targets is empty. Double check the makefile configuration and the build target.",
```
bobbrow, line 10: Just a suggestion. The loc team is inconsistent with punctuation across languages. I have a PR open on CMake Tools where I move the `'` into the substitution strings to avoid this.  I recommend the same here.

i18n/chs/src/tree.i18n.json
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+// Do not edit this file. It is machine generated.
+{
+   "tree.build.target": "Build target: [{0}]",
```
bobbrow, line 7: >[{0}] [](http://example.com/codeflow?start=39&length=5)

same suggestion here. I recommend you move the punctuation `[]` into the string being substituted.

package.nls.json
andreeis: Yes, I forgot about that, I was in the middle of editing when I stopped adding strings to push devDependencies in package.json to unblock the locBuild pipeline run. Fix now, submitting a PR update with all other strings required in package.(nls.)json.

i18n/chs/src/launch.i18n.json
andreeis: Ok, got it.

PR #270, opened 2022-01-31
Ignore only tests sources, not the whole out folder

andreeis:


bobbrow:
I think you might also need to add back in `node_modules` since we do not use webpack here.

bobbrow:
And by "add back in", I mean remove the exclusion for `node_modules`

andreeis:
Thank you for pointing this, I didn't know.

PR #278, opened 2022-02-08
fix makePath not properly resolved

### LINKED ISSUE #237: `makePath` does not work
On my system, `make` command is not defined - make and other compiler tools are not on PATH. 
It seems like `makefile.makePath` is not properly resolved. [Similar to this issue](https://github.com/microsoft/vscode-makefile-tools/issues/236), I've experimentally discovered, that:
* If `makefile.makePath` is defined in X.code-workspace `settings` group, absolute path to the `make.exe` is a valid settings. Leaving only the path to the folder where `make.exe` can be found, is not enough.
* Any setting in .vscode/settings.json didn't work for me. 

How this should work? Can the extension docs be a bit more precise about how user should configure these settings?

hagaigold:
add "make" when only a directory path was specified- fix #237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # What's New?
 
+## 
+Bug fixes:
+- makePath: add "make" when only a directory path was specified. [#237](https://github.com/microsoft/vscode-makefile-tools/issues/237)
+
 ## 0.3.1
 Bug fixes:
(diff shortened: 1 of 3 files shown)


andreeis (approved):

andreeis:
@hagaigold, thank you for fixing this issue. It will be included in the upcoming 0.4 release.

PR #280, opened 2022-02-14
Changes suggested by the loc team

andreeis:
@Colengms, this is what was recommended in https://dev.azure.com/ceapex/CEINTL/_workitems/edit/541630?src=WorkItemMention&src-action=artifact_link.

I noticed that CMT does not set isCreatePrSelected to true. Anything I should think about when changing this boolean?

### DIFF
--- a/jobs/loc/TranslationsImportExport.yml
+++ b/jobs/loc/TranslationsImportExport.yml
@@ -36,7 +36,7 @@ steps:
   inputs:
     locProj: 'jobs/loc/LocProject.json'
     outDir: '$(Build.ArtifactStagingDirectory)'
-    isCreatePrSelected: true
+    isCreatePrSelected: false
     prSourceBranchPrefix: 'locfiles'
     packageSourceAuth: 'patAuth'
     patVariable: '$(OneLocBuildPat)'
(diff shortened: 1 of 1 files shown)


jobs/loc/TranslationsImportExport.yml
```diff
@@ -36,11 +36,12 @@ steps:
   inputs:
     locProj: 'jobs/loc/LocProject.json'
     outDir: '$(Build.ArtifactStagingDirectory)'
-    isCreatePrSelected: false
```
Colengms, line 39: You will want to leave this field as `false` if you are using: https://github.com/microsoft/vscode-makefile-tools/blob/main/translations_auto_pr.js

In cpptools and cmaketools, we check in the i18n files (which are more straight-forward to manually maintain, if necessary) to the repo.  I believe if you were to use the PR they generate, that would skip the `translates-import` script, and it would instead try to check in `xlf` files.  If you want to use to the same approach as cpptools and cmaketools, I would leave this as `false`.

Colengms (changes requested):

Colengms (approved):

PR #288, opened 2022-02-25
Fix issue with missing translations

Colengms:
The gulp scripts copied from cpptools and CMake Tools were set up to work with webpack.  Makefile Tools does not use webpack, so some changes were needed.

vscode-nls depends on a "rewrite" of the calls to `localize` by vscode-nls-dev.  (This improves performance, as it switches to using indexes to look up loc strings instead of using key names.  I suspect there are other reasons it's needed.).  That rewrite was occurring for cpptools and CMake Tools during the webpack stage.  This change for Makefile Tools performs that rewrite explicitly in the gulp build task.

Because the rewrite step implicity requires converting the TypeScript code to JavaScript, the call to `tsc -p ./` was superfluous when running `compile-production`.  (It would actually replace the rewritten js files with non-rewritten contents).

Also, fixed the output location of the loc bundles.  `dist` was the location files were output for webpack.  Instead, I switched this to out, which is where it writes js files when building with `tsc`.

This fixed the issue for me when copying the files into an installed extension directory.  @andreeis , could you try building this into a VSIX to validate it fixes the issue?

### DIFF
--- a/package.json
+++ b/package.json
@@ -688,7 +688,7 @@
     "scripts": {
         "vscode:prepublish": "yarn compile-production",
         "compile": "yarn install && tsc -p ./",
-        "compile-production": "yarn install && yarn run translations-generate && yarn run compile",
+        "compile-production": "yarn install && yarn run translations-generate",
         "translations-export": "gulp translations-export",
         "translations-generate": "gulp translations-generate",
         "translations-import": "gulp translations-import",
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -206,12 +206,11 @@ const generatedSrcLocBundle = () => {
     return tsProject.src()
         .pipe(sourcemaps.init())
         .pipe(tsProject()).js
-        .pipe(nls.createMetaDataFiles())
+        .pipe(nls.rewriteLocalizeCalls())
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -688,7 +688,7 @@
     "scripts": {
         "vscode:prepublish": "yarn compile-production",
         "compile": "yarn install && tsc -p ./",
-        "compile-production": "yarn install && yarn run translations-generate && yarn run compile",
+        "compile-production": "yarn install && yarn run translations-generate",
```
andreeis, line 691: Removing the ending "&& yarn run compile" from the "compile-production" script, will not result in not compiling the project? I don't see it added somewhere else.

All else looks good and thank you very much for helping me figure out the whole localization functionality. I will test right now.

andreeis: Never mind... I just read your PR description again where you explain why you remove "yarn run compile".
The vsix I built from your branch doesn't work for me but maybe there's something else going on from my side, I'll figure out and let you know.

andreeis: I think it was a problem of installation on top of what was there. I uninstalled the Makefile extension and reapplied the vsix from your branch and all works great now, all strings. Thank you very much Colen. I'll merge this in.

andreeis (approved):

Issue #290, opened 2022-03-08
Propose removing debug/run options based on a property. To turn the extension into a pure building extension.


jdmchp:
* This issue exists to get feedback on whether a feature we would like to implement makes sense to your team. We'd implement the PR if the idea makes sense to you.
We would like to contribute a pull request to disable some GUI elements of the vscode-makefile-tools extension based on the existence of a property: makefile.enableDebugRunLocalOptions (shown in the File->Settings a Enable Local Feature [or any other better name]).
If this property is set to true (which would be the default) in the File->Settings, then the following elements would  be available on the GUI:
![image](https://user-images.githubusercontent.com/100438651/157526202-1990b0fa-40a1-4a81-87dd-5001c6defb92.png)
If the property is disabled:
![image](https://user-images.githubusercontent.com/100438651/157526871-29ad6026-94d3-45f1-8e1a-2305d4754c37.png)

* This is the problem we are trying to solve:
We find that having more than one way to issue a debug command 
![image](https://user-images.githubusercontent.com/100438651/157329539-9a29d902-5900-4a10-92e0-a20e7150ec01.png)
is confusing.
We are using the vscode-makefile-tools extension to build embedded targets (cross-compiling). We then use the [marus25.cortex-debug](https://marketplace.visualstudio.com/items?itemName=marus25.cortex-debug) to debug these embedded targets (elf files) in a target outside the host.

When we run both the vscode-makefile-tools extension and the marus25.cortex-debug, the user can may try to debug an elf file (s)he has created. The elf file can be debug via the marus25.cortex-debug via:
![image](https://user-images.githubusercontent.com/100438651/157330233-1e879cd8-d712-45c5-82fd-5c023961d1ac.png)
In this usage case, the option:
![image](https://user-images.githubusercontent.com/100438651/157330305-074792fa-fa3c-49ff-bcb9-53f1ab78a725.png)

does not apply.

PR #291, opened 2022-03-11
Proposed fix for microsoft/vscode-makefile-tools/issues/290

jdmchp:
As mentioned in [issue 290](https://github.com/microsoft/vscode-makefile-tools/issues/290), we would like to propose making the local (to the host) options of debugging and running optional. This makes the vscode-makefile-tools extension friendlier to use with other extensions that allow debugging in embedded systems.
making_local_run_debug_operations_optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -505,7 +505,7 @@
                     "description": "%makefile-tools.configuration.makefile.compileCommandsPath.description%",
                     "scope": "resource"
                 },
-                "makefile.optionalFeatures": {
+                "makefile.panel.visibility": {
                     "type": "object",
                     "default": null,
                     "properties": {
--- a/package.json
+++ b/package.json
@@ -502,6 +502,12 @@
                     "default": null,
                     "description": "%makefile-tools.configuration.makefile.compileCommandsPath.description%",
                     "scope": "resource"
+                },
+                "makefile.enableLocalDebugRun": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "%makefile-tools.configuration.makefile.enableLocalDebugRun.description%",
+                    "scope": "resource"
                 }
             }
         },
@@ -548,11 +554,11 @@
                 },
                 {
                     "command": "makefile.launchDebug",
-                    "when": "makefile:fullFeatureSet"
+                    "when": "makefile:localFeatureSet"
                 },
                 {
                     "command": "makefile.launchRun",
-                    "when": "makefile:fullFeatureSet"
+                    "when": "makefile:localFeatureSet"
                 },
                 {
                     "command": "makefile.setBuildConfiguration",
@@ -564,7 +570,7 @@
                 },
                 {
                     "command": "makefile.setLaunchConfiguration",
-                    "when": "makefile:fullFeatureSet"
+                    "when": "makefile:localFeatureSet"
                 },
                 {
                     "command": "makefile.outline.preConfigure",
@@ -580,11 +586,11 @@
                 },
                 {
                     "command": "makefile.outline.launchDebug",
-                    "when": "makefile:fullFeatureSet"
+                    "when": "makefile:localFeatureSet"
                 },
                 {
                     "command": "makefile.outline.launchRun",
-                    "when": "makefile:fullFeatureSet"
+                    "when": "makefile:localFeatureSet"
                 },
                 {
                     "command": "makefile.outline.setBuildConfiguration",
@@ -596,7 +602,7 @@
                 },
                 {
                     "command": "makefile.outline.setLaunchConfiguration",
-                    "when": "makefile:fullFeatureSet"
+                    "when": "makefile:localFeatureSet"
                 }
             ],
             "view/title": [
@@ -627,12 +633,12 @@
                 },
                 {
                     "command": "makefile.outline.launchDebug",
-                    "when": "view == makefile.outline",
+                    "when": "view == makefile.outline && makefile:localFeatureSet",
                     "group": "navigation@2"
                 },
                 {
                     "command": "makefile.outline.launchRun",
-                    "when": "view == makefile.outline",
+                    "when": "view == makefile.outline && makefile:localFeatureSet",
                     "group": "navigation@3"
                 }
             ],
--- a/package.json
+++ b/package.json
@@ -173,7 +173,8 @@
                 "icon": {
                     "light": "res/light/edit.svg",
                     "dark": "res/dark/edit.svg"
-                }
+                },
+                "enablement": "makefile:localFeatureSet"
             },
             {
                 "command": "makefile.outline.configure",
--- a/package.json
+++ b/package.json
@@ -141,7 +141,8 @@
                 "icon": {
                     "light": "res/light/debug.svg",
                     "dark": "res/dark/debug.svg"
-                }
+                },
+                "enablement": "makefile:localDebugFeature"
             },
             {
                 "command": "makefile.outline.launchRun",
@@ -149,7 +150,8 @@
                 "icon": {
                     "light": "res/light/run.svg",
                     "dark": "res/dark/run.svg"
-                }
+                },
+                "enablement": "makefile:localRunFeature"
             },
             {
                 "command": "makefile.outline.setBuildConfiguration",
@@ -173,8 +175,7 @@
                 "icon": {
                     "light": "res/light/edit.svg",
                     "dark": "res/dark/edit.svg"
-                },
-                "enablement": "makefile:localFeatureSet"
+                }
             },
             {
                 "command": "makefile.outline.configure",
@@ -504,11 +505,26 @@
                     "description": "%makefile-tools.configuration.makefile.compileCommandsPath.description%",
                     "scope": "resource"
                 },
-                "makefile.enableLocalDebugRun": {
-                    "type": "boolean",
-                    "default": true,
-                    "description": "%makefile-tools.configuration.makefile.enableLocalDebugRun.description%",
-                    "scope": "resource"
+                "makefile.panel.visibility": {
+                    "type": "array",
+                    "default": [],
+                    "items": 
+                    {
+                        "type": "object",
+                        "default": null,
+                        "properties": {
+                            "enableLocalDebugging": {
+                                "type": "boolean",
+                                "description": "%makefile-tools.configuration.makefile.panel.visibility.enableLocalDebugging.description%",
+                                "default": true
+                            },
+                            "enableLocalRunning": {
+                                "type": "boolean",
+                                "description": "%makefile-tools.configuration.makefile.panel.visibility.enableLocalRunning.description%",
+                                "default": true
+                            }
+                        }
+                    }
                 }
             }
         },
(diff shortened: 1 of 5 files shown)


src/tree.ts
```diff
@@ -103,6 +103,12 @@ export class LaunchTargetNode extends BaseNode {
             item.collapsibleState = vscode.TreeItemCollapsibleState.None;
             item.tooltip = localize("launch.target.currently.selected.for.debug.run.in.terminal",
                                     "The launch target currently selected for debug and run in terminal.\n{0}", this._toolTip);
+            // enablement in makefile.outline.setLaunchConfiguration is not
```
jdmchp, line 106: Trying set this `TreeItem` to be disabled via [enablement](https://github.com/jdmchp/vscode-makefile-tools/blob/main/package.json#L177) in the `command`: `makefile.outline.setLaunchConfiguration`

src/tree.ts
```diff
@@ -170,8 +176,12 @@ export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode>
         if (node) {
             return node.getChildren();
         }
-
-        return [this._currentConfigurationItem, this._currentBuildTargetItem, this._currentLaunchTargetItem];
+        if (configuration.getEnableLocalDebugRun()) {
```
jdmchp, line 179: Since `enablement` does not work. I had to use this if. Could this be related to [this issue](https://github.com/microsoft/vscode/issues/102794)?

src/configuration.ts
```diff
@@ -1278,6 +1295,15 @@ export async function initFromStateAndSettings(): Promise<void> {
...
+            subKey = "enableLocalDebugRun";
+            let updatedEnableLocalDebugRun : boolean | undefined = workspaceConfiguration.get<boolean>(subKey);
+            if (updatedEnableLocalDebugRun !== enableLocalDebugRun) {
+                readEnableLocalDebugRun();
+                readCurrentLaunchConfiguration();
+                extension.getState().configureDirty = true;
+                updatedSettingsSubkeys.push(subKey);
+            }
+
             // Final updates in some constructs that depend on more than one of the above settings.
```
andreeis, line 1307: I may be wrong, but I think you don't need configureDirty to be true here. Toggling between view or hide the UI elements doesn't need the extension to go through the reconfigure phase. In case you encountered issues which inspired you to write this (as opposed to looking at other similar code above which may have other reasons) let me know. It doesn't hurt but sometimes, for large projects, configure may be very long and we don't want to let it happen without being necessary.

Same question for calling readCurrentLaunchConfiguration. This doesn't take long, even one less reason to insist on removing it, but do you remember whether you needed that call?

src/tree.ts
andreeis: I am not familiar with that vscode bug but thanks for finding the connection and commenting on the alternative. I agree we need to do hide/show the tree element via an IF in the code, until the "enablement" keyword will be properly implemented in VSCode for tree items.

jdmchp: OK. I am glad you agree.

src/configuration.ts
jdmchp: >>as opposed to looking at other similar code above which may have other reasons)

Yes, basically I followed the pattern without understanding the implications. I will update the PR using your input.
Thanks a lot for the feedback.

src/tree.ts
```diff
@@ -103,6 +103,12 @@ export class LaunchTargetNode extends BaseNode {
             item.collapsibleState = vscode.TreeItemCollapsibleState.None;
             item.tooltip = localize("launch.target.currently.selected.for.debug.run.in.terminal",
                                     "The launch target currently selected for debug and run in terminal.\n{0}", this._toolTip);
+            // enablement in makefile.outline.setLaunchConfiguration is not
+            // disabling this TreeItem
+            item.command = {
+                command: "makefile.outline.setLaunchConfiguration",
+                title: "%makefile-tools.command.makefile.setLaunchConfiguration.title%"
+            }
```
andreeis, line 111: Until I try this myself to verify, you did not encounter any issues for using "title" and "%" in typescript this way, as opposed to how it's done in package.json? Is the text associated with %makefile-tools.command.makefile.setLaunchConfiguration.title% properly taken from package.nls.json?

andreeis:
@jdmchp, thank you for taking the time to implement this. I reviewed and left a few comments (to which you responded already). But the most significant suggestion I have is about the new setting. Having the experience of another similar extension based on cmake build technologies, which is a few years older, we ended up needing to show/hide more things, not just the debug UI elements. Instead of adding a new top level setting for each UI element that we may need to hide/show in future, let's follow that pattern of having all the visibility settings encapsulated in one object top level setting, to which we can add in time based on necessities.

The GitHub repository for the CMake Tools VSCode extension is here: https://github.com/microsoft/vscode-cmake-tools. You can follow in package.json and in the typescript sources how settings like "cmake.touchbar.visibility" or "cmake.statusbar.visibility" are defined and used. We don't have a "status bar" concept in the Makefile Tools extension and no "touch bar" special support yet so I imagine you can name the settings collection like "makefile.panel.visibility", with one entry for launch and in future, when somebody will need it, another entry to build (cmake tools extension needed this for debug and build, to hide/show them based on user scenarios that I don't know very well).

jdmchp:
@andreeis ,
>> let's follow that pattern of having all the visibility settings encapsulated in one object top level setting, to which we can add in time based on necessities.
>> 3 settings: one for debug, one for run and one for build. Currently you have debug/run together into one setting.

Got it. Let me look at the CMake Tools VSCode extension. The pattern of having a single central place to enable/disable functionality makes more sense.

Thanks for the guidance.

src/tree.ts
jdmchp: As far as I can tell, yes.

andreeis:
@jdmchp, I forgot to let you know earlier. The CI tests were not hanging because of changes in your PR. I just fixed the problem somewhere in the extension and the fix will get merged in soon.

jdmchp:
Hello @andreeis ,
I updated this PR so that multiple optional features can be enabled/disabled by the `makefile.panel.visibility` property. This property now has an array of optional features that can be toggled. Also, I removed any unnecessary operations (like setting the state's dirty flag). I added two features that can be toggled independently of each other "local debugging" and "local running". I did not make the build feature optional, since to me, the extension's whole point of being is to build. But that could be easily done. If a new feature option is needed to be enabled/disabled and `enablement` is used to control it, then a change to package.json and a change to [this object](https://github.com/jdmchp/vscode-makefile-tools/blob/main/src/configuration.ts#L112) will be all that is needed.

Thanks for you feedback.

pr_291_single_prop_controls_multiple_features

package.json
```diff
@@ -502,6 +504,27 @@
                     "default": null,
                     "description": "%makefile-tools.configuration.makefile.compileCommandsPath.description%",
                     "scope": "resource"
+                },
+                "makefile.panel.visibility": {
+                    "type": "array",
```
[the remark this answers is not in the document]
jdmchp: Hellp @bobbrow , thanks for the quick replay. My goal is to have a list of features that need to be toggled independently of each other. Being a list, I thought of using an array. The goal being able to add optional (toggle-able features) [easily](https://github.com/jdmchp/vscode-makefile-tools/blob/main/src/configuration.ts#L103). The original PR had a single optional feature (build/debug locally).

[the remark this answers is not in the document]
bobbrow: An object would be a better representation of what you want.  If you do an array of objects, then this would be valid syntax in the settings:
```
{
   "makefile.panel.visibility": [
    {
      "enableLocalDebugging": false
    },
    {
      "enableLocalDebugging": true
    }
  ]
}
```
It would be better to just do an object:
```
{
   "makefile.panel.visibility": {
     "enableLocalDebugging": false
   }
}
```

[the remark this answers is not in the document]
jdmchp: The use of an array made [iterating](https://github.com/jdmchp/vscode-makefile-tools/blob/main/src/configuration.ts#L138) through the possible different 'option flags' easy. I can refactor to make the `makefile.panel.visibility` and object and then iterate through its properties instead.
Thanks for the input.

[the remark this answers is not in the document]
jdmchp: I updated the PR. Using an object makes the code more readable. So, thanks for that suggestion.
pr_291_single_object_controls_multiple_features
I am new to typescript/json. Things flow differently when using java/xml (what I am used to).

package.json
```diff
@@ -502,6 +504,22 @@
                     "default": null,
                     "description": "%makefile-tools.configuration.makefile.compileCommandsPath.description%",
                     "scope": "resource"
+                },
+                "makefile.optionalFeatures": {
```
bobbrow, line 508: I think the old name `makefile.panel.visibility` was better.  "Optional features" is a bit too vague.  And if we copy the pattern from CMake Tools, the individual settings can have simpler names like "debug" and "run" because the context is already understood.

jdmchp: I see your point. Let me put that back. Do you think I should have a type in configurations.ts that matches the `makefile.panel.visibility` exactly:
```
// as described in makefile.panel.visibility
type MakefilePanelVisibility = {
    enableLocalDebugging: boolean,
    enableLocalRunning: boolean
};
```
I am not sure what the conventions are. But having an easily relatable name for the type that represents the object that is encapsulated by makefile.panel.visibility makes sense to me.

jdmchp: OK. Changed the property back to `makefile.panel.visibility` and adjusted types to be relatable to that name.

bobbrow: I'll let @andreeis comment on that. She knows the code better than I do. I tend to chime in more on the stuff that our users would notice like settings, commands, and such.

jdmchp: OK. Thanks.

jdmchp: @andreeis , if there is any more work needed on this PR, please let me know. Thanks in advance for your input.

andreeis:
@jdmchp, I am pushing a commit with the following changes I did:
- fixed some linter complaints in the new code (anything related with ; or spacing or empty lines, return types, etc...)
- renamed enableLocalDebugging to debug and enableLocalRunning to run (since name of the parent setting "makefile.panel.visibility" sets the context of these settings that now have smaller name)
- minor refactoring in tree.ts, since you added updateTree I just called it in other places where we were doing what updateTree does now
- fixed a bug in src/configuration.ts (updateOptionalFeaturesWithWorkspace) where I found a bug while testing, which I explain in the comment (we were not restoring defaults properly in some case)
- fixed a bug in the settings update mechanism (src/configuration.ts - initFromStateAndSettings - onDidChangeConfiguration). You forgot to push a new subKey for "makefile.panel.visibility" and we were not tracking the telemetry about settings changes, like for the others.

Additionally, the telemetry was broken for variables like "makefile.name1.name2.name3" etc... and I have PR https://github.com/microsoft/vscode-makefile-tools/pull/309 to fix that. It is separate from your PR.

jdmchp:
@andreeis 
Thanks for the clean up and the fixes for the 2 bugs (when `makefile.panel.visibility` is not completely defined, and not pushing the telemetry key). I updated my fork locally and tested too. Looks good.

andreeis (approved):

andreeis:
@jdmchp, thank you again for implementing this support. It is included in Makefile Tools 0.5.0 which was released today.

PR #301, opened 2022-03-31
Added the ability to name launch configurations

jlwoolf:
I posted an issue about 4 hours ago about how long path names were problematic for launch configurations. It annoyed me enough that I decided to go ahead and modify the extension so that launch configurations can be named. With this, if you name a configuration, it will appear as "NAME |>binaryPath(Args...)" and drop the cwd entirely. It will still use the cwd if specified for debugging and running, just changes how it is displayed. 

Actually, the way I have set this up should allow for a lot of customization as launch targets are now loaded as a map of type <LaunchConfigurations, String> and the names displayed for selection are created from the configuration. The selection then maps back to the original configuration, so, as a result, you can modify how an option is displayed versus the actual. Hopefully what I have described makes sense.


jlwoolf:
I realize my commit has some changes that don't quite relate to the implementation (mostly related to typing). I was bothered by the typescript errors so I added that. They can be ignored.

PR #306, opened 2022-04-13
Resolve ${workspaceRoot}/${workspaceFolder} in makePath

DepthDeluxe:
* Useful for providing environment wrapping scripts which bootstrap
  environment for make commands to be successful.

I created this change after spending a long time confused why this extension could not find my make wrapping script.  I needed to delegate `make` to this script so it could source a local development environment that is required to successfully build the project.

**Delegating Make Script**
```
#!/bin/bash

source .envrc &>/dev/null
make $@
```

These changes now allow for setting `makefile.makePath` to a value that is prefixed with `${workspaceFolder}`.  Confirmed via local testing that my wrapping scripts stored in `.vscode/make` now resolves properly.

I also updated the documentation to make these requirements clear because it was initially strange to me that this particular config would be treated differently than the others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,9 @@ The extension is going to invoke the script before every configure operation.
 ### Configuring your project
 By default, the extension will attempt to use a `make` program that resides within your $PATH to configure
 the project.  If you use a different flavor of the make tool or if it is not in your $PATH, use the
-`makefile.makePath` setting to instruct the extension where to find it.
+`makefile.makePath` setting to instruct the extension where to find it.  Provide a file/command that is in the
+system path, prefixed with `${workspaceRoot}`, or an absolute path as relative paths will not be resolved
+properly.
 
 The extension can also avoid running the `make` program when it configures your project, if you point the
 `makefile.buildLog` setting to the output of a build.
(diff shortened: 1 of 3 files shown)


andreeis:
@DepthDeluxe, thank you for implementing this new scenario. We plan to support variables expansion in the near future and before we consider taking a small change as a precursor of that wider scoped feature, let's see if you can achieve the same goal with support that is currently in the extension, like the preconfigure script.

Unless I miss something, I think that you can point the extension to the script that creates your needed environment, via "makefile.preConfigureScript" setting in .vscode/settings.json. Then set "makefile.alwaysPreConfigure": "true" in the same settings.json and this will cause the environment script to be run before Makefile Tools "configures" your code base. Once that happens, you have that environment available for building, running a target in the terminal or debugging it.

Try this and let me know how this works. Even if you switch to using the preconfigure script, we can still use parts of your PR when we get to implement the full feature of variables expansion. So still, very much appreciated.
If the preconfigure script doesn't help, let's see, maybe we fix some bugs in that feature or in the end, we are not against taking your change in now, if it turns out that your scenario cannot be addressed with the current functionality of the extension.

DepthDeluxe:
@andreeis, I confirmed setting your proposed configuration works great for me.  Thanks for the suggestion!

In that case at a minimum, I would recommend we document that `makefile.makePath` works differently from other settings in that it doesn't do cwd substitution.  Since the error message is a generic "make command not found", it makes it really unclear why the plugin can find other files but not this command.  I had to dig into the source code before realizing that.

andreeis (approved):

andreeis:
@DepthDeluxe, I actually merged this in, could still be useful as is, before we get to a more thorough approach of variables expansion.

PR #356, opened 2022-08-26
Additional settings updates fix

andreeis:
Don't write into settings when we cannot honor the user given output folder, specifically when the user changes "makefile.extensionOutputFolder" into something that doesn't exist.


bobbrow (approved):
This looks fine. What about the other 2 `.update()` calls?

andreeis:
> This looks fine. What about the other 2 `.update()` calls?

One is about "unnamed makefile configurations", where we add names ourselves (but the user already wrote "makefile.configurations" into settings.json so we can safely edit an improvement.
The other is about adding a new launch configuration and happens when the user selects one from the quick pick, so we have the user permission.

PR #409, opened 2023-01-17
Fix issues with compiler switch parsing and do command substitution for commands marked as safe

bobbrow:
We no longer attempt to emulate shell command parsing.  Instead we delegate that task to the shell itself.  Some interesting cases present themselves on *nix platforms where backticks are interpreted as inline commands to be run.  We can't trust all commands by default as that could be a security risk, so a new setting is also added to manage which commands are safe to run when we do the command line parsing.

My first iteration of a fix added a new parameter to `util.spawnChildProcess` which motivated me to consolidate the arguments into an object.  That change wasn't strictly necessary after a few iterations where the command execution was moved to a different part of the code.  However, I think the change adds value to the maintainability of the code, so I kept it.

Some formatting changes also happened to take place as part of this, but they are mostly whitespace related.  To simplify the review, please hide whitespace changes.

A future change may build on this and change `ToolInvocation.arguments` into an array of strings.  There are a few functions that still look through the string for specific switches using regular expressions and those regex's could potentially be removed if the arguments are already split into an array.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -463,6 +463,15 @@ export function readExcludeCompilerNames(): void {
     logger.message(`Exclude compiler names: ${excludeCompilerNames}`);
 }
 
+let safeCommands: string[] | undefined;
+export function getSafeCommands(): string[] | undefined { return safeCommands; }
+export function setSafeCommands(compilerNames: string[]): void { safeCommands = compilerNames; }
+export function readSafeCommands(): void {
+    let workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("makefile");
+    safeCommands = workspaceConfiguration.get<string[]>("safeCommands");
+    logger.message(`Exclude compiler names: ${safeCommands}`);
+}
+
 let dryrunSwitches: string[] | undefined;
 export function getDryrunSwitches(): string[] | undefined { return dryrunSwitches; }
 export function setDryrunSwitches(switches: string[]): void { dryrunSwitches = switches; }
@@ -1020,6 +1029,7 @@ export async function initFromStateAndSettings(): Promise<void> {
     readDryrunSwitches();
     readAdditionalCompilerNames();
     readExcludeCompilerNames();
+    readSafeCommands();
     readCurrentMakefileConfiguration();
     readMakefileConfigurations();
     readCurrentTarget();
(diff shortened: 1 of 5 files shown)


src/configuration.ts
```diff
@@ -463,6 +463,15 @@ export function readExcludeCompilerNames(): void {
     logger.message(`Exclude compiler names: ${excludeCompilerNames}`);
 }
 
+let safeCommands: string[] | undefined;
+export function getSafeCommands(): string[] | undefined { return safeCommands; }
+export function setSafeCommands(compilerNames: string[]): void { safeCommands = compilerNames; }
+export function readSafeCommands(): void {
+    let workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("makefile");
+    safeCommands = workspaceConfiguration.get<string[]>("safeCommands");
+    logger.message(`Exclude compiler names: ${safeCommands}`);
```
andreeis, line 472: Copy paste typo: "Exclude compiler names" --> "Safe commands"

src/util.ts
```diff
@@ -319,6 +329,40 @@ export function spawnChildProcess(
...
+            x = x.replace(match[0], `\`${match[1]} ${match[2]}\``);
+        }
+        match = regex.exec(x);
+    }
+    return x;
+}
+
+// Run a shell command and return the output.
+async function runCommand(x: string, opt: child_process.SpawnOptions): Promise<string> {
+    const child = child_process.spawn('/bin/bash', ['-c', `"${x}"`], opt);
```
andreeis, line 351: Here we need also the windows equivalent for "/bin/bash -c" which is "cmd -c". I'm doing the changes I am noticing as I review and test and will post a PR commit follow-up.

andreeis: My mistake, backtick does not apply on windows. A bailout to not run bin/bash on windows intead.

src/util.ts
```diff
@@ -655,11 +699,13 @@ export function thisExtensionPackage(): PackageJSON {
     const pkg: PackageJSON = thisExtension().packageJSON as PackageJSON;
 
     return {
-      name: pkg.name,
-      publisher: pkg.publisher,
-      version: pkg.version,
-      contributes: pkg.contributes
+        name: pkg.name,
```
gcampbell-msft, line 702: Nit: indentation seems changed, is this expected / wanted?

src/util.ts
```diff
@@ -623,22 +667,22 @@ export function scheduleTask<T>(task: () => T): Promise<T> {
...
-       setImmediate(async () => {
-           try {
-               const result: T = await task();
-               resolve(result);
-           } catch (e) {
-               reject(e);
-           }
-       });
-   });
+    return new Promise<T>((resolve, reject) => {
```
gcampbell-msft, line 670: Nit: Indentation

gcampbell-msft:
Overall, I don't see any issues with this PR. However, I'll default to @andreeis for when we want to wrap this up and merge, and then include in a release.

andreeis:
> Overall, I don't see any issues with this PR. However, I'll default to @andreeis for when we want to wrap this up and merge, and then include in a release.

I have several changes applied on top of this and in the end I stopped when found some aspects that required more testing. I'll update this PR later with what I found and I believe we should prioritize for 0.8 (didn't create the project yet).

src/util.ts
bobbrow: yes. Should be 4 spaces

src/util.ts
bobbrow: `replaceCommands` should exit early on Windows.

gcampbell-msft:
This looks like if we adjust this and bring it up to date, it would be able to implement / fix #670

Issue #341, opened 2022-07-09
The value of miDebuggerPath is invalid in Ubuntu


karriszou:
**Operating System:** _Ubuntu 20.04 LTS_
**output of type 'code -v'**:
_1.68.1
30d9c6cd9483b2cc586687151bcbcd635f373630
x64_

There is a strange issue: everytime i restart Ubuntu and press the **Debug button of the makefile-tool** to debug the program, an error message appears: **Unable to start debugging. The value of miDebuggerPath is invalid**

This is my settings.json:
**{
    "makefile.extensionOutputFolder": "./.vscode",
    "makefile.makePath": "/usr/bin/make",
    "makefile.configurations": [
        {
            "name": "mydebug",
            "makeArgs": []
        }
    ],
    "makefile.launchConfigurations": [
        {
            "miDebuggerPath": "/usr/bin/gdb",
            "MIMode": "gdb",
            "stopAtEntry": false,
            "cwd": "/home/code/ioq3/build/debug-linux-x86_64",
            "binaryPath": "/home/code/ioq3/build/debug-linux-x86_64/ioquake3.x86_64",
            "binaryArgs": [
                "+map", "q3dm3", "+set", "developer", "1", "+set", "sv_pure", "0", "+set", "vm_game", "0", "+set", "vm_cgame", "0", "+set", "vm_ui", "0"
            ]
        }
    ],
    "C_Cpp.default.configurationProvider": "ms-vscode.makefile-tools",
}**

the makefile-tool 'Configuration' option is "mydebug" not "Default".
**the outpt of type 'which gdb'**: _/usr/bin/gdb_, the same as "miDebuggerPath" option.

The first two times I set the miDebuggerPath option to "", **and there is no lauch.json and tasks.josn file**, somehow I can debug the program from makefile-tool (not F5). But when I restart Ubuntu for the third time and debug program the error always appears no matter if I set miDebuggerPath to "/usr/bin/gdb" or "", i don't know if this is a bug or a configure error.

How can I debug my program from makefile-tool (not F5), can someone help me?

karriszou:
This time i set the 'miDebuggerPath' to "/usr/bin", and it debug program successful. I don't know why...

karriszou:
[...]

PR #428, opened 2023-03-08
update commands and miDebuggerPath resolution

gcampbell-msft:
There were commands that were not exposed in package.json but were in activation events, so I fixed that.

Additionally, I added checks that should allow for the usage of direct file paths in `miDebuggerPath`, rather than only supporting folder directories.


package.json
```diff
@@ -99,6 +99,38 @@
                 "command": "makefile.launchRun",
                 "title": "%makefile-tools.command.makefile.launchRun.title%"
             },
+            {
```
bobbrow, line 102: These commands are intentionally omitted from `package.json` so that they do not appear in the Command Palette. These commands look like the one we exposed for Command Substitution.  They actually don't need to be declared in `package.json`.

andreeis: I was about to comment about the same thing. Otherwise, the change in launch.ts looks good.

gcampbell-msft: The changes for those commands have been reverted. 👍

andreeis (approved):

gcampbell-msft:
I should probably reference the bug that this is fixing. #341

src/extension.ts
```diff
@@ -233,6 +233,8 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
         await launcher.runCurrentTarget();
     }));
 
+
+    /** Start of commands that shouldn't be exposed in package.json */
```
andreeis, line 237: Since you added this comment, you can also specify besides "shouldn't be exposed in package.json" that these are command substitution (used by VSCode in launch.json and tasks.json).

gcampbell-msft: Good idea. I'll add.

andreeis (approved):

PR #430, opened 2023-03-09
if the make file path is a directory, search for the file

gcampbell-msft:
This checks if the provided makefile path from the user settings is a directory. If it's a directory, search for `Makefile` and `makefile` inside that directory. If we find the makefile, set the `makefileUsed` to that path. 

@andreeis While investigating this I wondered if we would benefit from consolidating `makefile.makefilePath` and `makefile.makeDirectory`. Maybe we could consolidate and then detect if we've gotten the directory or the file name and use that to determine whether to use `-f` of `-C`? Just thinking out loud.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -690,10 +690,15 @@ export function getCommandForConfiguration(configuration: string | undefined): v
     if (makefileUsed) {
         // check if the makefile path is a directory. If so, try adding `Makefile` or `makefile` 
         if (util.checkDirectoryExistsSync(makefileUsed)) {
-            makefileUsed = path.join(makefileUsed, "Makefile");
-            if (!util.checkFileExistsSync(makefileUsed)) {
-                makefileUsed = path.join(makefileUsed, "makefile");
+            let makeFileTest = path.join(makefileUsed, "Makefile");
+            if (!util.checkFileExistsSync(makeFileTest)) {
+                makeFileTest = path.join(makefileUsed, "makefile");
             }
+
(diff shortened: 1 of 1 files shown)


andreeis:
@gcampbell-msft , thank you for the PR, it looks good, I'll approve it.
Good question. The naming of the two made me also think of combining them. But they may be totally different at times. It happens that -f and -C point towards the same folder but they may be both defined and point to different places. -f points to the root makefile that "make" build tool parses first (and then includes anything else necessary) while -C refers to what folder the activity happens in (equivalent of CD or pushd so that all relative paths are resolved to that).
So, it's not a rule that the folder given with "makefile.makeDirectory" (which generates -C) should be the same as the folder that contains the makefile given with "makefile.makefilePath" (which generates -f).

gcampbell-msft:
> @gcampbell-msft , thank you for the PR, it looks good, I'll approve it. Good question. The naming of the two made me also think of combining them. But they may be totally different at times. It happens that -f and -C point towards the same folder but they may be both defined and point to different places. -f points to the root makefile that "make" build tool parses first (and then includes anything else necessary) while -C refers to what folder the activity happens in (equivalent of CD or pushd so that all relative paths are resolved to that). So, it's not a rule that the folder given with "makefile.makeDirectory" (which generates -C) should be the same as the folder that contains the makefile given with "makefile.makefilePath" (which generates -f).

That makes sense, thanks for the explanation. Tt seems like my fix is probably the way to go for now, so I'll go ahead and merge. Thanks!

Issue #417, opened 2023-02-13
[Windows] Ampersand in path


KrestenHelstrup:
Hi,

The extension does not correctly resolve paths with ampersand on windows systems.
Would it be possible to fix that in a future release ?

andreeis:
@KrestenHelstrup, thank you for pointing this out. While investigating this, the fix turned to be quick so it will be included in the 0.7 release.

KrestenHelstrup:
@andreeis - Sounds good. When is the expected release date of 0.7.0 ?

andreeis:
@KrestenHelstrup, a fix for this will be released with 0.7 next week but if you would like to see how this works earlier you can install one vsix from [this link](https://github.com/microsoft/vscode-makefile-tools/actions/runs/4451312188).

go to the Extensions panel on the left
click "..."
run "Install from vsix" from the context menu

Let us know if you encounter any other issues or if you have any other suggestions/recommendations in this area of functionality.

gcampbell-msft:
@KrestenHelstrup Due to 0.7.0 being available with this fix, we are closing this issue, please re-open if you hit issues.

Issue #416, opened 2023-02-10
Case sensitive on Windows


KrestenHelstrup:
hi,
Looks like the plugin is case sensitive even on windows systems which are not case sensitive. In my makefile I have something like:
C_SRC +=  \
../somefolder/somefile.c
../Main.c
The proejct compiles without any issues but intellisense is not working as expected. In the extension log I see:
Configuration for file xxx\someFolder\somefile.c was not found. CppTools will set a default configuration.
As you see the folder is someFolder with capital F and if I change the makefile to correct casing everything works as expected.

Would it be possible for the plugin to be case-insensitive on windows systems ?
Issues like these are quite hard to debug.

Thanks in advance

andreeis:
@KrestenHelstrup , thank you for pointing this out. We have a fix which will be included in the 0.7 release.

andreeis:
@KrestenHelstrup, a fix for this will be released with 0.7 next week but if you would like to see how this works earlier you can install one vsix from [this link](https://github.com/microsoft/vscode-makefile-tools/actions/runs/4451312188).

go to the Extensions panel on the left
click "..."
run "Install from vsix" from the context menu

Let us know if you encounter any other issues or if you have any other suggestions/recommendations in this area of functionality.

gcampbell-msft:
Due to 0.7.0 being available with this fix, we are closing this issue, please re-open if you hit issues.

CherryDT:
"on Windows" is an incorrect oversimplification!

As I pointed out in a comment on the PR (just wasn't sure which place was better to write this), whether case-sensitivity is enabled depends not only on the platform but also on the filesystem used (one may have mounted a Linux filesystem) and **even on which folder the file is in** because on NTFS you can enable case-sensitivity per folder (I have it enabled on my projects folder, for example, because I want to catch capitalization errors early before they crash on a Linux server later on)!

PR #431, opened 2023-03-11
Fixes for 0.7

andreeis:
This PR addresses the following:
- change in the extension launch.json, debugging was broken (did not identify since when and by what but recently) because typescript sources were not found
- I found that "makefile.*" settings updates were not handled properly on edit/update for the scenario of "add folder" to workspace because the listener was also expecting only .vscode/settings.json. I believe it is correct to not assume anything about the file path of the settings because VSCode settings reader knows to read from everywhere and to apply the proper rules of defaults and overrides. No work item opened for this. Fix is in configuration.ts
- https://github.com/microsoft/vscode-makefile-tools/issues/416: for windows, when we identify a source file for which we want to save its configuration in the table, we can uppercase its path so that later, when CppTools asks about it, whatever casing it chooses, we'll find it. Fix in extension.ts and cpptools.ts
- https://github.com/microsoft/vscode-makefile-tools/issues/417: make sure paths with & are also quoted, not only the once with spaces. Fix in util.ts

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -36,7 +36,7 @@
             "stopOnEntry": false,
             "sourceMaps": true,
             "outFiles": [
-                "${workspaceFolder}/dist/**/*.js"
+                "${workspaceFolder}/out/**/*.js"
             ],
             "preLaunchTask": "compile",
         },
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -1115,8 +1115,7 @@ export async function initFromStateAndSettings(): Promise<void> {
     // More than one setting may be updated on one settings.json save,
     // so make sure to OR the dirty state when it's calculated by a formula (not a simple TRUE value).
     vscode.workspace.onDidChangeConfiguration(async e => {
-        if (vscode.workspace.workspaceFolders &&
-            e.affectsConfiguration('makefile', vscode.workspace.workspaceFolders[0].uri)) {
+        if (vscode.workspace.workspaceFolders && e.affectsConfiguration('makefile')) {
             // We are interested in updating only some relevant properties.
             // A subset of these should also trigger an IntelliSense config provider update.
             // Avoid unnecessary updates (for example, when settings are modified via the extension quickPick).
--- a/src/cpptools.ts
+++ b/src/cpptools.ts
@@ -28,7 +28,10 @@ export class CppConfigurationProvider implements cpp.CustomConfigurationProvider
     private workspaceBrowseConfiguration: cpp.WorkspaceBrowseConfiguration = { browsePath: [] };
 
     private getConfiguration(uri: vscode.Uri): SourceFileConfigurationItem | undefined {
-        const norm_path: string = path.normalize(uri.fsPath);
+      let norm_path: string = path.normalize(uri.fsPath);
+      if (process.platform === "win32") {
+         norm_path = norm_path.toUpperCase();
+      }
 
         // First look in the file index computed during the last configure.
         // If nothing is found and there is a configure running right now,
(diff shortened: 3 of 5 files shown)


.vscode/launch.json
```diff
@@ -36,7 +36,7 @@
             "stopOnEntry": false,
             "sourceMaps": true,
             "outFiles": [
-                "${workspaceFolder}/dist/**/*.js"
+                "${workspaceFolder}/out/**/*.js"
```
gcampbell-msft, line 39: Another way to fix this would be to let the debugger just search the entire workspace, this would be done by simply removing `outFiles`. This is what I did to be able to hit breakpoints. As long as it works I'm okay with just switching the path, but I thought I'd mention it.

src/configuration.ts
```diff
@@ -1115,8 +1115,7 @@ export async function initFromStateAndSettings(): Promise<void> {
     // More than one setting may be updated on one settings.json save,
     // so make sure to OR the dirty state when it's calculated by a formula (not a simple TRUE value).
     vscode.workspace.onDidChangeConfiguration(async e => {
-        if (vscode.workspace.workspaceFolders &&
-            e.affectsConfiguration('makefile', vscode.workspace.workspaceFolders[0].uri)) {
+        if (vscode.workspace.workspaceFolders && e.affectsConfiguration('makefile')) {
```
gcampbell-msft, line 1118: What is the significance of removing the `vscode.workspace.workspaceFolders[0].uri` argument?

src/cpptools.ts
```diff
@@ -28,7 +28,10 @@ export class CppConfigurationProvider implements cpp.CustomConfigurationProvider
     private workspaceBrowseConfiguration: cpp.WorkspaceBrowseConfiguration = { browsePath: [] };
 
     private getConfiguration(uri: vscode.Uri): SourceFileConfigurationItem | undefined {
-        const norm_path: string = path.normalize(uri.fsPath);
+      let norm_path: string = path.normalize(uri.fsPath);
```
gcampbell-msft, line 31: Nit: The spacing/tabbing looks off by a couple of spaces, it looks like it's not indented enough.

gcampbell-msft:
I left a couple of comments, and of course I'm still ramping up, but it looks good/reasonable to me. 👍

gcampbell-msft (approved):

.vscode/launch.json
andreeis: Thanks for mentioning it, indeed that might be more clear. I only followed the example in the .vscode/launch.json of cpptools and cmake tools extensions. I wonder if there are any consequences from removing "outFiles"... I'll keep this in mind for an eventual later update

src/configuration.ts
andreeis: When I wrote that code I wanted to avoid all the below analysis when other files were modified and the code path hit than the .vscode/settings.json of that project. But I overlooked the scenario of user settings (defined in another settings.json) and workspace settings (defined in a totally different file, without even the format "settings.json"). VSCode reads from everywhere and knows to apply the rules of defaults/overrides... From wherever it reads, we should only care that the settings of Makefile Tools are impacted and trust the value we're given when reading. I'm sure this is the root cauuse of many bugs.

src/cpptools.ts
andreeis: thx for spotting that. fixed with latest commit.

src/cpptools.ts
```diff
@@ -28,7 +28,10 @@ export class CppConfigurationProvider implements cpp.CustomConfigurationProvider
     private workspaceBrowseConfiguration: cpp.WorkspaceBrowseConfiguration = { browsePath: [] };
 
     private getConfiguration(uri: vscode.Uri): SourceFileConfigurationItem | undefined {
-        const norm_path: string = path.normalize(uri.fsPath);
+        let norm_path: string = path.normalize(uri.fsPath);
+        if (process.platform === "win32") {
+           norm_path = norm_path.toUpperCase();
+        }
```
CherryDT, line 34: This is an incorrect assumption. Whether case-sensitivity is enabled depends not only on the platform but also on the filesystem used (one may have mounted a Linux filesystem) and **even on which folder the file is in** because on NTFS you can enable case-sensitivity per folder (I have it enabled on my projects folder, for example, because I want to catch capitalization errors early before they crash on a Linux server later on)!

PR #432, opened 2023-03-11
Report not found stuff in Makefile Tools UI

andreeis:
Addressing the whole group of complaints around the popups for Make and/or Makefile missing, this PR does the following:

- removes the remaining "Make missing" popup (the Makefile missing popup was removed already with a PR from the community https://github.com/microsoft/vscode-makefile-tools/pull/415)
- keeps the logging of the missing files as it was, in the makefile tools output channel at the same time as it happened before (at load time, when reading and analyzing settings, before the configure step)
- makes the Makefile Tools UI visible also when a Makefile is missing (which before it was hidden by design), but other actions or features or UI elements remain hidden when no makefile is found (this is for the scenario of users who have code bases that contain makefiles somewhere - maybe tests or ... something else not relevant for us - but their project is not makefile based and they don't want any messages or impact from our extension being activated)
- adds 3 nodes in the UI with information about the make build tool, the root makefile that was identified and the build log (from start, we did not think of including build log in this analysis... but, when a build log is given, makefile and make are allowed to be missing so it is more clear to show info about all 3 pieces). The information given in the UI via title of tree node and tooltip is the winning rule from settings (either a default or an override...) and whether it is missing or not. The action associated with the edit pen icon is to open the settings, because there the user can define a different path for these 3: make, makefile or build log.
- I found more cases of "check if file exists" APIs failing when paths were correct and files present, because of quotes. Those quotes would exist because of "tab-tab" in windows console when adding more to the PATH variable (the OS adds those quotes when it identifies a space in the folder that's being discovered with "tab"). I would later open VSCode from that environment and our extension functions like "find in path" would split the path variable based in ";", quotes would remain and somehow the file checks would fail because of that. We found scenarios like these in the past as well but now I moved the fix in a more central place for more coverage. Fix in util.ts. No work item opened specifically for this but I'm sure it is related to some of the reports of "Makefile UI not seen" even when the makefile was present, reports that I was not able to reproduce on my end.
-

### DIFF
--- a/package.json
+++ b/package.json
@@ -178,6 +178,14 @@
                 }
             },
             {
+               "command": "workbench.action.openWorkspaceSettings",
+               "title": "%workbench.action.openWorkspaceSettings.title%",
+               "icon": {
+                   "light": "res/light/edit.svg",
+                   "dark": "res/dark/edit.svg"
+               }
+            },
+            {
                 "command": "makefile.outline.configure",
                 "title": "%makefile-tools.command.makefile.configure.title%",
                 "icon": {
@@ -537,7 +545,6 @@
             "makefile__viewContainer": [
                 {
                     "id": "makefile.outline",
-                    "when": "makefile:fullFeatureSet",
                     "name": "%makefile-tools.configuration.views.makefile.outline.description%"
                 }
             ]
@@ -615,6 +622,10 @@
                 {
                     "command": "makefile.outline.setLaunchConfiguration",
                     "when": "makefile:localDebugFeature"
+                },
+                {
+                  "command": "workbench.action.openWorkspaceSettings",
+                  "when": "always"
                 }
             ],
             "view/title": [
@@ -635,22 +646,22 @@
                 },
                 {
                     "command": "makefile.outline.buildTarget",
-                    "when": "view == makefile.outline",
+                    "when": "makefile:fullFeatureSet && view == makefile.outline",
                     "group": "navigation@1"
                 },
                 {
                     "command": "makefile.outline.buildCleanTarget",
-                    "when": "view == makefile.outline",
+                    "when": "makefile:fullFeatureSet && view == makefile.outline",
                     "group": "1_makefileOutline@4"
                 },
                 {
                     "command": "makefile.outline.launchDebug",
-                    "when": "view == makefile.outline && makefile:localDebugFeature",
+                    "when": "makefile:fullFeatureSet && view == makefile.outline && makefile:localDebugFeature",
                     "group": "navigation@2"
                 },
                 {
                     "command": "makefile.outline.launchRun",
-                    "when": "view == makefile.outline && makefile:localRunFeature",
+                    "when": "makefile:fullFeatureSet && view == makefile.outline && makefile:localRunFeature",
                     "group": "navigation@3"
                 }
             ],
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -635,22 +646,22 @@
                 },
                 {
                     "command": "makefile.outline.buildTarget",
-                    "when": "view == makefile.outline",
+                    "when": "makefile:fullFeatureSet && view == makefile.outline",
```
andreeis, line 649: The following adds of "makefile:fullFeatureSet" are needed now because the UI is visible now even if fullFeatureSet is false, while before it was not visible, so no extra reason to hide all its components if the whole panel was hidden.

package.nls.json
```diff
@@ -8,6 +8,7 @@
    "makefile-tools.command.makefile.setBuildConfiguration.title": "Makefile: Set the current build configuration",
    "makefile-tools.command.makefile.setBuildTarget.title": "Makefile: Set the target to be built by make",
    "makefile-tools.command.makefile.setLaunchConfiguration.title": "Makefile: Set the make launch configuration",
+   "workbench.action.openWorkspaceSettings.title": "Preferences: Open Workspace Settings",
```
andreeis, line 11: Not sure how to properly localize this string... it's different than all the rest, it doesn't belong to makefile tools, but how do I reference a string localize-able from another extension or from vscode itself?

src/configuration.ts
```diff
@@ -740,7 +747,23 @@ export function getCommandForConfiguration(configuration: string | undefined): v
         logger.message(`Deduced command '${configurationMakeCommand} ${configurationMakeArgs.join(" ")}' for configuration "${configuration}"`);
     }
 
-    // Validation and warnings about properly defining the makefile and make tool.
+    // Check for makefile path on disk: we search first for any makefile specified via the makefilePath setting,
```
andreeis, line 750: This whole green block I moved from below, it was by mistake under an IF about not having a build log provided. Having a build log given via "makefile.buildLog" only bypasses the "make" invocation when we calculate IntelliSense but when users want to build, we still need to know where make and makefile are.

src/configuration.ts
```diff
@@ -1402,9 +1414,10 @@ export async function initFromStateAndSettings(): Promise<void> {
             }
 
             // Final updates in some constructs that depend on more than one of the above settings.
```
andreeis, line 1416: configureDirty is about we run or not configure later when querying the right info.
In here, we need to re-analyze configure params and update the tree nodes (the ones about make/makefile/build.log, not about configuration/build-target/launch-target, those get updated via their SET actions... but make/makefile/build.log don't have yet SET actions, they are modified via settings updates).

src/util.ts
```diff
@@ -150,12 +156,7 @@ export function toolPathInEnv(name: string): string | undefined {
     // (the concept of kit for cmake extension)
 
     return envPathSplit.find(p => {
-        let fullPath: string = path.join(p, path.basename(name));
```
andreeis, line 153: If we would remove quotes before we call checkFileExistsSync here, we are not covered when we call checkFileExistsSync from somewhere else where we don't verify the same thing.
Best to remove these quotes right inside checkFileExistsSync, before the stat API call.

package.nls.json
gcampbell-msft: Good question. I'm not sure I know of a way to do this. 

Would it be bad to simply localize it for ourselves here? Presumably the strings will be the same.

package.json
gcampbell-msft: To make sure I understand this right: 
The goal is to make the UI visible in all cases. However, there are a certain subset of features that will not be visible or available unless `fullFeatureSet` is set to `true`?

gcampbell-msft: A follow up question: What is the purpose of the UI being visible now even if `fullFeatureSet` is false? Would there be buttons or things that are now disabled due to `fullFeatureSet` being false?

andreeis: The UI needs to be visible now in order to see the status (make or makefile not found). Otherwise this approach (a status in this UI) would be useless because the same cause for the status to be reported used to cause the UI to hide.
So now the UI is visible but all buttons that before didn't need to hide, are not visible now.
Another reason to make sense for the UI to be accessible is that the UI is one entry point for correcting the problem that caused the undesired situation: there are edit buttons (pen icons) to open the settings and change "makefile.makePath" and "makefile.makefilePath" settings.

PR #438, opened 2023-03-20
Dependabot fixes

gcampbell-msft:
This PR should points minimatch to at least 3.0.5 which is the lowest patched value.

Additionally, in order to make this easier, I updated some dependencies, which made some changes in tsconfig.json and the code to satisfy the types.

@andreeis, Please make sure that the code type changes I made, specifically the change in src/make.ts where I made a decision to avoid an undefined value of e.exitCode

### DIFF
--- a/src/make.ts
+++ b/src/make.ts
@@ -353,7 +353,7 @@ export async function doBuildTarget(progress: vscode.Progress<{}>, target: strin
             let disposable: vscode.Disposable = vscode.tasks.onDidEndTaskProcess((e: vscode.TaskProcessEndEvent) => {
                 if (e.execution.task.name === makefileBuildTaskName) {
                     disposable.dispose();
-                    resolve(e.exitCode ?? ConfigureBuildReturnCodeTypes.cancelled);
+                    resolve(e.exitCode ?? ConfigureBuildReturnCodeTypes.other);
                 }
             });
         }));
--- a/src/make.ts
+++ b/src/make.ts
@@ -350,10 +350,10 @@ export async function doBuildTarget(progress: vscode.Progress<{}>, target: strin
         await vscode.tasks.executeTask(myTask);
 
         const result: number = await(new Promise<number>(resolve => {
-            let disposable: vscode.Disposable = vscode.tasks.onDidEndTaskProcess(e => {
+            let disposable: vscode.Disposable = vscode.tasks.onDidEndTaskProcess((e: vscode.TaskProcessEndEvent) => {
                 if (e.execution.task.name === makefileBuildTaskName) {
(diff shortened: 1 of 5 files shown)


src/make.ts
```diff
@@ -350,10 +350,10 @@ export async function doBuildTarget(progress: vscode.Progress<{}>, target: strin
         await vscode.tasks.executeTask(myTask);
 
         const result: number = await(new Promise<number>(resolve => {
-            let disposable: vscode.Disposable = vscode.tasks.onDidEndTaskProcess(e => {
+            let disposable: vscode.Disposable = vscode.tasks.onDidEndTaskProcess((e: vscode.TaskProcessEndEvent) => {
                 if (e.execution.task.name === makefileBuildTaskName) {
                     disposable.dispose();
-                    resolve(e.exitCode);
+                    resolve(e.exitCode ?? ConfigureBuildReturnCodeTypes.cancelled);
```
andreeis, line 356: Did you actually try this? You cancelled the build via the bottom-right popup button, that caused the task to return undefined and so we now ensure we return cancelled?

tsconfig.json
```diff
@@ -1,22 +1,23 @@
...
+		"strict": true,   /* enable all strict type-checking options */
 		/* Additional Checks */
 		// "noImplicitReturns": true, /* Report error when not all code paths in function return a value. */
 		// "noFallthroughCasesInSwitch": true, /* Report errors for fallthrough cases in switch statement. */
-		// "noUnusedParameters": true,  /* Report errors on unused parameters. */
+		// "noUnusedParameters": true,  /* Report errors on unused parameters. */,
 	},
 	"exclude": [
-		"node_modules",
+		"node_modules/**",
```
andreeis, line 20: Looks good, but also for my learning experience what inspired this. Squiggle recommended or what happened. CMake Tools may need the same.

gcampbell-msft: Good question, I did this to ensure that building failures didn't happen from node_modules, but I realized that the typescript dependency needed to be updated, this change can likely be reverted.

gcampbell-msft: Just tested and this isn't needed, reverting.

src/make.ts
gcampbell-msft: Followed up offline, changing to `other` due to the exitCode being 0 when the task is cancelled with the UI, and the handling of a cancelled task is done by the boolean `cancelBuild`, thus there isn't a need for having that here. I also tested and no telemetry gets duplicated either way. My expectation is that for all normal cases of a user trying to build, is that the `exitCode` will always be defined.

andreeis (approved):

Issue #441, opened 2023-03-22
"Build Target" list does not support single character targets


nick-hebi:
This is a bug that is related to the regex on [Line 55 of parser.ts](https://github.com/microsoft/vscode-makefile-tools/blob/main/src/parser.ts#L55). The extension fails to recognize build targets in the Makefile that are a single character long (such as "A:").

I believe that this regex should be changed from `"^(?!\\n?[#\\.])(?<!^\\n?# Not a target:\\s*)\\s*(\\S+[^:]):\\s+"` to `"^(?!\\n?[#\\.])(?<!^\\n?# Not a target:\\s*)\\s*(\\S*[^:]):\\s+"`.

This regex is checking for a group of characters that ends in a single ':' character. The usage of `(\\S+[^:])` fails on single character build target cases such as "A:" because the `+` consumes the first character and causes the `[^:]` condition to fail. Changing the `+` to a `*` remedies this issue while preserving the intended behavior of this line.

nick-hebi:
PR Opened https://github.com/microsoft/vscode-makefile-tools/pull/442

andreeis:
@nick-hebi , thank you for finding this issue and for submitting a fix. Being so small we can easily include in the 0.7 release.

PR #442, opened 2023-03-22
Update regex in parser.ts to remedy issue #441

nick-hebi:
https://github.com/microsoft/vscode-makefile-tools/issues/441

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -52,7 +52,7 @@ export async function parseTargets(cancel: vscode.CancellationToken, verboseLog:
         // Skip lines starting with {#,.} or preceeded by "# Not a target" and extract the target.
         // Additionally, if makefile.phonyOnlyTargets is true, include only targets
         // succeeded by "#  Phony target (prerequisite of .PHONY).".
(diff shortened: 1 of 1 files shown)


gcampbell-msft:
@nick-hebi This is a nit, but could you leave a comment, similar to the comment from your issue you posted, that explains the need for the change? It could be useful for context for any future development.

nick-hebi:
Copied from the [issue](https://github.com/microsoft/vscode-makefile-tools/issues/441) for visibility:

This is a bug that is related to the regex on [Line 55 of parser.ts](https://github.com/microsoft/vscode-makefile-tools/blob/main/src/parser.ts#L55). The extension fails to recognize build targets in the Makefile that are a single character long (such as "A:").

I believe that this regex should be changed from `"^(?!\\n?[#\\.])(?<!^\\n?# Not a target:\\s*)\\s*(\\S+[^:]):\\s+"` to `"^(?!\\n?[#\\.])(?<!^\\n?# Not a target:\\s*)\\s*(\\S*[^:]):\\s+"`.

This regex is checking for a group of characters that ends in a single ':' character. The usage of `(\\S+[^:])` fails on single character build target cases such as "A:" because the `+` consumes the first character and causes the `[^:]` condition to fail. Changing the `+` to a `*` remedies this issue while preserving the intended behavior of this line.

gcampbell-msft (approved):

PR #445, opened 2023-03-24
Ensure UI is able to display correct not found versus found status

gcampbell-msft:
During testing, I found a bug where in Windows, if the user had `make` on path, it wasn't finding it and it was displaying `(not found)` in the UI, this fixes it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     },
     "license": "SEE LICENSE IN LICENSE.txt",
     "engines": {
-        "vscode": "^1.74.0"
+        "vscode": "^1.76.0"
     },
     "bugs": {
         "url": "https://github.com/Microsoft/vscode-makefile-tools/issues",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     },
     "license": "SEE LICENSE IN LICENSE.txt",
     "engines": {
-        "vscode": "^1.76.0"
+        "vscode": "^1.74.0"
     },
     "bugs": {
         "url": "https://github.com/Microsoft/vscode-makefile-tools/issues",
@@ -744,7 +744,7 @@
     "devDependencies": {
         "@types/mocha": "^9.0.0",
         "@types/node": "^10.17.17",
-        "@types/vscode": "^1.76.0",
+        "@types/vscode": "^1.74.0",
         "tslint": "^5.20.1",
         "tslint-microsoft-contrib": "^6.2.0",
         "tslint-no-unused-expression-chai": "^0.1.4",
--- a/src/tree.ts
+++ b/src/tree.ts
@@ -305,12 +305,13 @@ export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode>
        if (!pathInSettings) {
          return `${kind}: [Unset]`;
        }
-
-       const pathBase: string | undefined = (searchInPath && path.parse(pathInSettings).dir === "") ? path.parse(pathInSettings).base : undefined;
+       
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -744,7 +744,7 @@
     "devDependencies": {
         "@types/mocha": "^9.0.0",
         "@types/node": "^10.17.17",
-        "@types/vscode": "^1.76.0",
+        "@types/vscode": "^1.74.0",
         "tslint": "^5.20.1",
```
andreeis, line 748: Lower version??

src/tree.ts
```diff
@@ -305,12 +305,13 @@ export class ProjectOutlineProvider implements vscode.TreeDataProvider<BaseNode>
        if (!pathInSettings) {
          return `${kind}: [Unset]`;
        }
-
-       const pathBase: string | undefined = (searchInPath && path.parse(pathInSettings).dir === "") ? path.parse(pathInSettings).base : undefined;
+       
+       const pathInSettingsToTest = process.platform === "win32" && !pathInSettings.endsWith(".exe") && kind === "Make" ? pathInSettings.concat(".exe") : pathInSettings;
+       const pathBase: string | undefined = (searchInPath && path.parse(pathInSettingsToTest).dir === "") ? path.parse(pathInSettingsToTest).base : undefined;
        const pathInEnv: string | undefined = pathBase ? (path.join(util.toolPathInEnv(pathBase) || "", pathBase)) : undefined;
```
andreeis, line 311: Minor: I know I missed some linter setting somewhere. VSCode squiggles for me that it wants all the types for variables so that's why we define all the types when we declare. Let's keep pathInSettingsToTest consistent (or until we decide we don't want to do this and set VSCode to not complain about that).

andreeis:
Except the version which I asked why lower, the rest looks good. I don't know why the windows tests failed, by reading the logs it didn't seem caused by the PR so I am hoping it is a fluke in the infrastructure. Rerunning now, hopefully will pass. If not, we'll investigate on Monday.

andreeis (approved):

package.json
drok: I am reviewing the history of external dependencies, and it looks like this is a ticking 🧨 .

When dependencies are eventually mass updated, won't the breakage caused by 1.76 return, resulting in a regression? What can be done to guard against this expected future breakage? Surely hope alone is not enough.

Based on the merged `yarn.lock` it appears [that version 1.76.0 was resolved as a correct alternative to 1.74.0](https://github.com/microsoft/vscode-makefile-tools/commit/5b4d318518d4bb2a54ec90af791b904ef079d9bc#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de), and that is what's being used for the release package:

```diff
diff --git a/yarn.lock b/yarn.lock
index 2b7784e..f32eedc 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -231,7 +231,7 @@
   resolved "https://registry.yarnpkg.com/@types/node/-/node-10.17.60.tgz#35f3d6213daed95da7f0f73e75bcc6980e90597b"
   integrity sha512-F0KIgDJfy2nA3zMLmWGKxcH2ZVEtCZXHHdOQs2gSaQ27+lNeEfGxzkIw90aXswATX7AZ33tahPbzy6KAfUreVw==

-"@types/vscode@^1.76.0":
+"@types/vscode@^1.74.0":
   version "1.76.0"
   resolved "https://registry.yarnpkg.com/@types/vscode/-/vscode-1.76.0.tgz#967c0fbe09921818bbf201f1cbcb81b981c6249c"
   integrity sha512-CQcY3+Fe5hNewHnOEAVYj4dd1do/QHliXaknAEYSXx2KEHUzFibDZSKptCon+HPgK55xx20pR+PBJjf0MomnBA==
```

PR #457, opened 2023-04-12
Variable expansion - initial support

andreeis:
Variable expansion. Will add more explanations and update linux test baselines later. Have to fix merge conflicts first.


gcampbell-msft:
>     default: boolean;

We might want to use a name other than `default`, in some contexts it's a keyword and I worry it's possible to introduce hard to track bugs.

---
Refers to: src/configuration.ts:100 in ccb3bbc. [](commit_id = ccb3bbcb41b585b6c9f720a6d0a608d86b44a2b5, deletion_comment = False)

gcampbell-msft:
>                 "makefile.panel.visibility": {

Maybe this is just ignorance on my part, but I don't understand what this is for / what it does, an explanation here might help me understand some of the changes in `configuration.ts`

---
Refers to: package.json:518 in ccb3bbc. [](commit_id = ccb3bbcb41b585b6c9f720a6d0a608d86b44a2b5, deletion_comment = False)

src/configuration.ts
```diff
@@ -189,16 +187,10 @@ export function setMakePath(path: string): void { makePath = path; }
...
 // It represents a default to look for if no other path is already included
 // in "makefile.configurations.makePath".
-function readMakePath(): void {
-    let workspaceConfiguration: vscode.WorkspaceConfiguration = vscode.workspace.getConfiguration("makefile");
-    makePath = workspaceConfiguration.get<string>("makePath");
+async function readMakePath(): Promise<void> {
+    makePath = await util.getExpandedSetting<string>("makePath");
     if (!makePath) {
         logger.message("No path to the make tool is defined in the settings file.");
-    } else {
```
gcampbell-msft, line 197: Why is this part removed? Was it refactored to somewhere else?

gcampbell-msft (approved):
:shipit:

src/util.ts
```diff
@@ -603,9 +604,246 @@ export function resolvePathToRoot(relPath: string): string {
...
+//         and any commands.
+// TODO: Currently, after applying any expansion pattern, if the result is another expansion pattern
+// we log an error but in future let's handle the recursivity and complications of expanding anything
+// coming via this entrypoint.
+export async function expandVariablesInSetting(settingId: string, settingVal: string): Promise<string> {
+    // Try the predefined VSCode variable first. The regexp for ${variable} won't fit the others because of the ":".
+    let expandedSetting: string = settingVal;
+    let regexpVSCodeVar: RegExp = /(\$\{(\w+)\})|(\$\{(\w+):(.+?)\})/mg;
+    let result: RegExpExecArray | null = regexpVSCodeVar.exec(expandedSetting);
+    while (result) {
```
gcampbell-msft, line 721: >while (result) { [](http://example.com/codeflow?start=4&length=16)

Is this a while loop for future use of recursion?

src/configuration.ts
andreeis: We call resolvePathToRoot (which needs a different name and will change with a future refactoring... this makes a full path if the given string is rel path and we assume relative to root) after all strings that we read as paths, except makePath. For make, before this PR, we had resolveSubstitutedPath which was only replacing easy case of ${workspaceFolder} or ${workspaceRoot}, but if given a relative path we were not making it full path resolved to root, because if given "make" we want to also search in path. When we see an intentional path (abs or rel) for "make", we don't pull it from the path anymore. Calling resolvePathToRoot on "make" or "make.exe" will not benefit of the functionality of searching for make in path (look at util.ts toolPathInEnv and how we call it).

src/configuration.ts
```diff
@@ -335,8 +316,9 @@ export function readExtensionOutputFolder(): void {
     if (extensionOutputFolder) {
         if (!util.checkDirectoryExistsSync(extensionOutputFolder)) {
             if (!util.createDirectorySync(extensionOutputFolder)) {
+                extensionOutputFolder = extension.extensionContext.storagePath;
```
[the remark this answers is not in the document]
andreeis: Yes. And if user does not override extensionOutputFolder with a different location, it's a bit hard to know where to look for those files. What I do is look in the logs, we print where create dryrun, cache, etc.... then I go to that path from the output channel. We may want to create a command to help the user go to that location easier.

andreeis:
> > ```
> >             "makefile.panel.visibility": {
> > ```
> 
> Maybe this is just ignorance on my part, but I don't understand what this is for / what it does, an explanation here might help me understand some of the changes in `configuration.ts`
> 
> Refers to: package.json:518 in [ccb3bbc](https://github.com/microsoft/vscode-makefile-tools/commit/ccb3bbcb41b585b6c9f720a6d0a608d86b44a2b5). [](commit_id = [ccb3bbc](https://github.com/microsoft/vscode-makefile-tools/commit/ccb3bbcb41b585b6c9f720a6d0a608d86b44a2b5), deletion_comment = False)

The panel visibility was implemented by the community. See PR https://github.com/microsoft/vscode-makefile-tools/pull/291 and issue https://github.com/microsoft/vscode-makefile-tools/issues/290.

It implements the ability to make various extension features optional and hide them from the UI, starting with debug and run. More features can be added to this visible/hidden toggle functionality when needed.

Regarding type MakefilePanelVisibilityDescription and the default boolean property... Let me think... We could rename it to defaultVisibility or defaultValue.... but I've seen many types with properties called "default", I wonder if you can give an example of issue that can happen? Maybe "default: " as in a switch statement versus a property in an object, or?

gcampbell-msft (approved):

PR #462, opened 2023-04-21
Improve UI options

gcampbell-msft:
This PR enhances and adds abilities to the Makefile UI. 

1. This PR updates the "edit" capability in the UI to navigate directly to the relevant UI setting, rather than simply to the settings page.
2. This PR adds an option for makefile and build log to also open the file. 
    1. This also tries to stay smart about whether the file is present, and if it's not, the option to open the file goes away.

Gif:
EnhancedUI


src/extension.ts
```diff
@@ -357,6 +366,43 @@ export async function activate(context: vscode.ExtensionContext): Promise<void>
     configuration.initFromState();
     await configuration.initFromSettings(true);
 
+    context.subscriptions.push(vscode.commands.registerCommand('makefile.outline.openMakefilePathSetting', () => {
+        return vscode.commands.executeCommand("workbench.action.openSettings", "makefile.makefilePath");
```
andreeis, line 370: I suggest "openWorkspaceSettings" instead of "openSettings". For all occurrences.

gcampbell-msft: Done! 👍

andreeis (approved):

PR #484, opened 2023-07-07
Implement a post configure script (sibling to pre-configure script)

gcampbell-msft:
This PR includes an update that adds support for a post-configure script. 

Additionally, it includes a slight refactor that allows for better sharing of the pre/post-configure script commands.


src/configuration.ts
```diff
@@ -13,6 +13,7 @@ import * as path from 'path';
 import * as telemetry from './telemetry';
 
 import * as nls from 'vscode-nls';
+import { log } from 'console';
```
andreeis, line 16: I didn't bring the PR over yet for testing, still just in reading mode, to really see if this is grey or not... but it doesn't look like you're using it, correct? Unused import?

package.json
```diff
@@ -125,6 +127,10 @@
                 "command": "makefile.preConfigure",
                 "title": "%makefile-tools.command.makefile.preConfigure.title%"
             },
+            {
+                "command": "makefile.postConfigure",
+                "title": "%makefile-tools.command.makefile.postConfigure.title"
```
andreeis, line 132: % missing at the end of title.

andreeis (approved):

package.json
gcampbell-msft: Good catch, fixing.

src/configuration.ts
gcampbell-msft: Good catch, fixing.

gcampbell-msft:
> I tested this and works well. So just my 2 comments, the % one being important. I see the test failures, expected baseline updates caused by one more setting we added. Not sure if we do any testing about preconfigure, probably nice to add something, some other time also good.

Do you see any locations that would cause failures in test other than the expected not printing post-configure setting? It seems that there is also some additional output that wasn't there before.

Issue #489, opened 2023-07-17
Makefile build/tests for Pull Requests are failing/hanging


gcampbell-msft:
All PR's are currently blocked on failing tests and hanging build/test.

PR #491, opened 2023-07-20
Dev/andris/makefile tools/fixes

andreeis:
This PR addresses the blocking bug https://github.com/microsoft/vscode-makefile-tools/issues/489, along with a few minor issues I noticed around commands in the palette.

Currently all PRs are blocked because while running tests we cannot trigger a quickPick and wait for user selection. This is as a hang for the tests infrastructure. So far, we've been avoiding opening quickPicks in test mode but now with several code changes and a combination of factors (related to variable expansion and reading settings more times to propagate updates plus something else we did not identify, because this did not reproduce when the feature was merged in) we got a new such situation.

One test is setting launch configurations, which usually writes into settings.json but we avoided so far writing updates into the settings.json of the tests to keep cleanup simple. Since now (after variable expansion) we read settings more times than before (because some values depend on each other and we have a 2 pass evaluation) we detect a launch configuration that is not present in settings.json, where usually for production code we would (correctly) unset the launch configuration. We could either not unset the launch config in such scenario for tests... or we could start writing into settings.json and ensure proper cleanup and deterministic outcomes in case tests stop in the middle. Both approaches end up having some ugly hacks being needed but I chose the latest because it appeared to have the fewest cons. 

Additionally, writing into settings.json triggers some settings updates events which lead to read settings again. This caused lots of extra logging about reading settings. I updated some baselines to account for that but where it was possible to identify a logging to be redundant we skipped over writing the message.

I'll write more comments now on the PR directly near the lines that need some extra explanations.


package.json
```diff
@@ -166,27 +166,27 @@
             },
             {
                "command": "makefile.outline.openMakefilePathSetting",
-               "title": "Open Makefile Path Setting",
+               "title": "%makefile-tools.command.makefile.openMakefilePathSetting.title%",
```
andreeis, line 169: All changes related to "title" and %% (accompanied by a line in package.nls.json) are to ensure we localize the name of these actions in the palette. There were also some typos from copy paste.

package.json
```diff
@@ -592,37 +592,45 @@
                     "command": "makefile.setLaunchConfiguration",
                     "when": "makefile:localDebugFeature"
                 },
+                {
+                    "command": "makefile.outline.configure",
+                    "when": "never"
+                },
```
andreeis, line 598: All the changes with "never" will cause the "outline" commands counterparts to not be visible in the palette. The palette was polluted with all these "outline" commands which make sense for the left side bar UI not for the palette.
I saw in the CMake Tools commands being written with "never" condition and I didn't understand why that was needed, thinking that removing the whole block would be the same as having it with "never". It's clear now how to make a command not show up in the palette.

src/configuration.ts
```diff
@@ -393,6 +393,7 @@ export function setConfigurationCachePath(path: string): void { configurationCac
 // and when the settings watcher detects a change of any properties that may impact the dryrun output.
 export async function readConfigurationCachePath(): Promise<void> {
     // how to get default from package.json to avoid problem with 'undefined' type?
+    let oldConfigurationCachePath = configurationCachePath;
```
andreeis, line 396: We need to do this in more places but so far I only focused on what was most annoying in the tests baseline-run diffs. Also, a rewrite/refactoring of the settings will fix many of these problems about the values, non expanded in settings, expanded in memory, out of date versus modified, type validation ...etc... CMake Tools has a great model for settings classes in typescript that we can follow.

src/configuration.ts
```diff
@@ -1507,7 +1500,7 @@ export async function setNewConfiguration(): Promise<void> {
 
         if (configureAfterCommand) {
             logger.message("Automatically reconfiguring the project after a makefile configuration change.");
-            await make.configure(make.TriggeredBy.configureAfterConfigurationChange);
+            await make.cleanConfigure(make.TriggeredBy.configureAfterConfigurationChange);
```
andreeis, line 1503: Setting a new configuration or a new target requires a clean config versus a non clean config, basically --always-make is needed to recalculate what compilation would look like for all the files even for the up to date ones. Otherwise, the user will see incorrect IntelliSense that slowly updates itself in time as they edit more source files.

src/configuration.ts
```diff
@@ -393,6 +393,7 @@ export function setConfigurationCachePath(path: string): void { configurationCac
 // and when the settings watcher detects a change of any properties that may impact the dryrun output.
 export async function readConfigurationCachePath(): Promise<void> {
     // how to get default from package.json to avoid problem with 'undefined' type?
```
gcampbell-msft, line 395: Is this comment stale?

src/configuration.ts
```diff
@@ -1243,23 +1246,13 @@ export async function initFromSettings(activation: boolean = false): Promise<voi
...
-                // If there is a directory defined within the configuration cache path,
-                // honor it and don't append to extensionOutputFolder.
-                let parsePath: path.ParsedPath = path.parse(updatedConfigurationCachePath);
-                if (extensionOutputFolder && !parsePath.dir) {
-                    updatedConfigurationCachePath = path.join(extensionOutputFolder, updatedConfigurationCachePath);
-                } else {
-                    updatedConfigurationCachePath = util.resolvePathToRoot(updatedConfigurationCachePath);
-                }
-            }
-            if (updatedConfigurationCachePath !== configurationCachePath) {
```
gcampbell-msft, line 1257: Is there some duplicated code here? It seems like we do similar checks in both this section and the `readConfigurationCachePath` method as well, could they be shared or combined?

src/configuration.ts
gcampbell-msft: Seems reasonable enough. We should keep an eye though on the issues list if people notice this causing performance hits.

src/configuration.ts
```diff
@@ -1608,7 +1601,7 @@ export async function selectTarget(): Promise<void> {
         if (configureAfterCommand) {
             // The set of build targets remains the same even if the current target has changed
             logger.message("Automatically reconfiguring the project after a build target change.");
-            await make.configure(make.TriggeredBy.configureAfterTargetChange, false);
+            await make.cleanConfigure(make.TriggeredBy.configureAfterTargetChange, false);
```
gcampbell-msft, line 1604: I assume this modification is due to the same reason as here: https://github.com/microsoft/vscode-makefile-tools/pull/491/files#r1269821024?

src/configuration.ts
```diff
@@ -1635,10 +1628,7 @@ export async function setLaunchConfigurationByName(launchConfigurationName: stri
             launchConfigAsInSettings.push(currentLaunchConfiguration);
             // Push into the processed 'in-memory' launch configurations array as well.
             launchConfigurations.push(currentLaunchConfiguration);
-            // Avoid updating the launchConfigurations array in settings.json for regression tests.
-            if (process.env['MAKEFILE_TOOLS_TESTING'] !== '1') {
```
gcampbell-msft, line 1639: If we're removing this, unless there are other places we check for it, should we remove the MAKEFILE_TOOLS_TESTING env variable? 

Not sure I'm necessarily suggesting this, just asking. It might be just as wise to leave it in case we need it in the future.

gcampbell-msft (approved):
Everything seems like reasonable changes to me. Thanks for the helpful comments!

src/configuration.ts
andreeis: I don't remember what I meant by that. Thanks for pointing it out. I removed the 6 occurrences of it.

src/configuration.ts
andreeis: There was duplication before these diffs, the red part. I'm not sure I see duplication now. Similar changes might be needed for other fragments up and down but I did not focus on other settings to cleanup. Let me know if this is what you meant.

src/configuration.ts
andreeis: yes

src/configuration.ts
andreeis: We still use it to disable other things in testing mode.

PR #494, opened 2023-07-28
localize strings that are in the progress bar

gcampbell-msft:
Add localization to strings that appear in the progress bar.


gcampbell-msft:
@andreeis ping for review.

Issue #151, opened 2021-04-28
Preconfigure arguments


andreeis:
Implement a new string (or string array) setting "preConfigureArgs" to be passed to the preconfigure script when it is run.
Global ("makefile.preConfigureArgs") and per configuration ("makefile.configurations[].preConfigureArgs").
The current workaround is to wrap a different script on top of the desired one, with the arguments written in the wrapper. This can quickly explode to many configuration scripts and wrappers, while it can be cleanly solved from the settings with a single preconfigure script.

andreeis:
This would be different than https://github.com/microsoft/vscode-makefile-tools/issues/144 and similar in implementation with https://github.com/microsoft/vscode-makefile-tools/issues/122.

156898293:
Is "makefile.preConfigureArgs" supposed to work in pre-release v0.8.16 (vscode 1.84.2)?
Because it is not.
Feel free to ignore this comment, its just a friendly notification.

156898293:
ok, will create a bug report

gcampbell-msft:
Assuming I'm right, I think I know your issue. Would this fix it? #530

156898293:
> Assuming I'm right, I think I know your issue. Would this fix it? #530
I don't think so... I created #531
If you need we can do a debug session over teams.

PR #495, opened 2023-07-31
Add pre/post configure script arguments globally and per configuration

gcampbell-msft:
See title. 

This will allow users to specify per configuration pre and post configure script arguments, as well as global. This will allow for people to use one central pre/post configure script across configurations as well as post versus pre and determine different functionality based on the parameters. 

Implements #151


gcampbell-msft:
@andreeis ping for review.

package.json
```diff
@@ -293,6 +293,22 @@
...
+                            },
+                            "postConfigureArgs": {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "default": [],
+                                "description": "%makefile-tools.configuration.makefile.postConfigureArgs.description%"
                             }
                         }
```
andreeis, line 313: Missing pre/post configure script at configuration level. We have args at configuration level and script+args at global level but I don't see script at config level.

src/configuration.ts
```diff
@@ -650,16 +678,26 @@ let configurationBuildLog: string | undefined;
...
+export function setConfigurationPostConfigureArgs(args: string[]): void { configurationPostConfigureArgs = args; }
+
 // Analyze the settings of the current makefile configuration and the global workspace settings,
 // according to various merging rules and decide what make command and build log
 // apply to the current makefile configuration.
 async function analyzeConfigureParams(): Promise<void> {
     getBuildLogForConfiguration(currentMakefileConfiguration);
     await getCommandForConfiguration(currentMakefileConfiguration);
     getProblemMatchersForConfiguration(currentMakefileConfiguration);
+    getPreConfigureArgsForConfiguration(currentMakefileConfiguration);
```
andreeis, line 696: Didn't test and debug yet, will understand better soon. But why have the getters of pre/post config args here? analyzeConfigParams are about configure phase (which invokes make --dry-run plus other switches which get calculated in here) and not about pre or post configure which doesn't even call "make --dry-run" but whatever is in the pre/post configure script. Eventually, maybe we need something like "analyzePrePostConfigureParams".

src/configuration.ts
```diff
@@ -866,6 +904,28 @@ export function getBuildLogForConfiguration(configuration: string | undefined):
     }
 }
 
+export function getPreConfigureArgsForConfiguration(configuration: string | undefined): void {
+    let makefileConfiguration: MakefileConfiguration | undefined = getMakefileConfiguration(configuration);
+    const preConfigArgs = makefileConfiguration?.preConfigureArgs;
```
andreeis, line 909: Too late (or not the right moment) to rename preConfigureArgs into globalPreConfigureArgs.... to be obviously different than an eventual localPreConfigureArgs, so if we leave preConfigureArgs alone (and we know those are the global ones) then let's rename preConfigArgs into localPreConfigureArgs. Otherwise, Config versus Configure in the two names is a not clear what is the difference and we'd have to read the code again to understand which is what.

src/configuration.ts
```diff
@@ -866,6 +904,28 @@ export function getBuildLogForConfiguration(configuration: string | undefined):
...
+    } else {
+        configurationPreConfigureArgs = preConfigureArgs;
+    }
+}
+
+export function getPostConfigureArgsForConfiguration(configuration: string | undefined): void {
+    let makefileConfiguration: MakefileConfiguration | undefined = getMakefileConfiguration(configuration);
+    const postConfigArgs = makefileConfiguration?.postConfigureArgs;
+
+    if (postConfigArgs) {
```
andreeis, line 922: same postConfigArgs renaming localPostConfigureArgs, same reason as for Pre above.

andreeis:
Didn't play (test, debug) with it yet but as visual review I commented about all that I noticed. Nothing major and looks almost ready.

package.json
gcampbell-msft: This was a design decision I made, but I'm happy to discuss whether it was the right one. 

My thoughts were that, this would allow users to have pre/post configure scripts at the global level, and they can use arguments to allow it to be per configuration. My estimation was that people would rather do that then to have an actual per configuration script.

src/configuration.ts
gcampbell-msft: This was done to ensure that the preConfigure and postConfigure settings are correctly established. I viewed these as directly related to configure, since it's directly tied to before configure and after configure.

src/configuration.ts
gcampbell-msft: Done.

src/configuration.ts
gcampbell-msft: 👍 Done.

package.json
andreeis: Oh, that is an alternative indeed. I just started from the idea of having per config scripts as well but I can see how having one script and per config args can work the same. If we find in future that users actually have different content of commands and script paragraphs, we can still add it later without breaking this design and the users that will start using it this way. I'm good as it is.

andreeis (approved):

PR #502, opened 2023-09-15
add workflow on releases that closes fixed-pending-release issues

gcampbell-msft:
This PR adds a workflow that runs when a new release is published and it goes through and closes issues that are labeled as `fixed-pending-release`.

### DIFF
--- a/.github/workflows/close-fixed-pending-release.yml
+++ b/.github/workflows/close-fixed-pending-release.yml
@@ -8,7 +8,7 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             - name: Close issues marked 'fixed-pending-release'
-              uses: gcampbell-msft/fixed-pending-release@0.0.11
+              uses: gcampbell-msft/fixed-pending-release@0.0.12
               with:
                 token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


andreeis (approved):

andreeis:
Super cool. How about another one, for "needs more info" older than a set time to write a comment to the customer whether they have additional info and phase 2, after another set time with no response close, with the comment to reopen as soon as they have the info to provide.

gcampbell-msft:
> Super cool. How about another one, for "needs more info" older than a set time to write a comment to the customer whether they have additional info and phase 2, after another set time with no response close, with the comment to reopen as soon as they have the info to provide.

This is a good idea, there are existing bots that can do this, we should chat with @snehara99 to try and mimic what is done in the CMake Tools repository.

snehara99:
> > Super cool. How about another one, for "needs more info" older than a set time to write a comment to the customer whether they have additional info and phase 2, after another set time with no response close, with the comment to reopen as soon as they have the info to provide.
> 
> This is a good idea, there are existing bots that can do this, we should chat with @snehara99 to try and mimic what is done in the CMake Tools repository.

definitely can chat about it :)

PR #511, opened 2023-10-04
fix changelog

gcampbell-msft:


gcampbell-msft:
Doesn't affect tests, only a changelog update.

PR #528, opened 2023-11-17
update changelog

gcampbell-msft:
Updating the CHANGELOG.md so that we track our feature for workspace trust.


gcampbell-msft:
Forcing merge because this won't affect tests.

PR #530, opened 2023-11-20
there was a bug where the first argument wasn't adding a space in front

gcampbell-msft:
see title.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # What's New?
 
 ## 0.8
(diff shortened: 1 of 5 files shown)


benmcmorran (approved):
You might want to split this into two PRs, but I'm not blocking on that.

gcampbell-msft:
> You might want to split this into two PRs, but I'm not blocking on that.

This is a good reminder, I'll leave it as one but if anything regresses this may be the one to take a look at. Thanks.

PR #559, opened 2024-02-07
Update to sign files

gcampbell-msft:


andreeis (approved):
Version should be 0.7 instead of 1.7, otherwise looks good. Thank you for this improvement.

PR #573, opened 2024-03-18
Remove dry-run popup, replace with configureOnOpen questions

gcampbell-msft:
Most importantly, this removes the problematic and frustrating to users dry-run pop-up. 

When you open a workspace that has a Makefile, you will first be prompted with this pop-up. 

image

Then, the makefile extension will start the configure process, and also give you this pop-up that will control what level we will save your choice in, either globally or only in this workspace. 
image

However, if you say no to the first pop-up, the configure process will not be started, and then you will be prompted with this:

image

This will allow you to either globally turn off configuring on open, or only for the current workspace.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -751,7 +751,7 @@ export async function activate(
       const chosen = await vscode.window.showInformationMessage<Choice1>(
         localize(
           "extension.configureOnOpen",
-          "Would you like to configure with the Makefile Tools extension?"
+          "Would you like to configure this workspace with the Makefile Tools extension?"
         ),
         {},
         { title: localize("yes", "Yes"), doConfigure: true },
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -732,13 +732,91 @@ export async function activate(
       )
     );
   }
-  // === Commands only for testing ===
+  // === End of commands only for testing ===
 
   const parseCompilerArgsScript: string = util.parseCompilerArgsScriptFile();
 
   // The extension VSIX stripped the executable bit, so we need to set it.
   // 0x755 means rwxr-xr-x (read and execute for everyone, write for owner).
   await fs.chmod(parseCompilerArgsScript, 0o755);
+
+  if (extension.getFullFeatureSet()) {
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -732,17 +731,105 @@ export async function activate(
...
   const parseCompilerArgsScript: string = util.parseCompilerArgsScriptFile();
 
   // The extension VSIX stripped the executable bit, so we need to set it.
   // 0x755 means rwxr-xr-x (read and execute for everyone, write for owner).
   await fs.chmod(parseCompilerArgsScript, 0o755);
 
-  if (configuration.getConfigureOnOpen() && extension.getFullFeatureSet()) {
-    // Always clean configure on open
-    await make.cleanConfigure(make.TriggeredBy.cleanConfigureOnOpen);
+  // TODO: Need to add logging and/or telemetry.
```
elizabethmorrow, line 742: Nit: I see you emitting telemetry here is this comment still relevant? 

Not a blocker, can be removed later.

src/telemetry.ts
```diff
@@ -422,3 +422,14 @@ function getPackageInfo(): IPackageInfo {
...
   };
 }
+
+export type ConfigureOnOpenScope = "user" | "workspace";
+export function logConfigureOnOpenTelemetry(
+  configureOnOpen: boolean,
+  scope: ConfigureOnOpenScope | undefined = undefined
+) {
+  logEvent("configureOnOpenPopup", {
+    configureOnOpen: configureOnOpen.toString(),
```
elizabethmorrow, line 432: Would this emit "[object null]" if configureonOpenPopup is never assigned?

src/extension.ts
gcampbell-msft: Ah, good point, I can remove this, telemetry has been added.

src/telemetry.ts
gcampbell-msft: I'm not sure what you mean? configureOnOpenPopup is simply the name of the event, so it will always emit an event and it will tag along those properties of `configureOnOpen` and `scope`.

elizabethmorrow (approved):

PR #629, opened 2024-07-29
update localization key

gcampbell-msft:


gcampbell-msft:
Not waiting for approval to get telemetry key updated sooner.

PR #649, opened 2024-09-13
ignore .js in codesign validation

gcampbell-msft:


jobs/release/prerelease.yml
```diff
@@ -29,6 +29,9 @@ extends:
             templateContext:
               type: releaseJob
               isProduction: true
+              sdl:
+                codeSignValidation:
+                  additionalTargetsGlobPattern: -|*.js
```
jophippe, line 34: I think you should use the **\\*.js pattern instead of *.js, since that's the pattern that Code Sign Validation uses to include .js files by default

jobs/release/release.yml
```diff
@@ -61,6 +61,9 @@ extends:
             templateContext:
               type: releaseJob
               isProduction: true
+              sdl:
+                codeSignValidation:
+                  additionalTargetsGlobPattern: -|*.json
```
jophippe, line 66: I think you meant .js here

gcampbell-msft: Resolved.

jobs/release/prerelease.yml
gcampbell-msft: Updated

PR #654, opened 2024-09-20
Ensure signing/packaging works as expected

gcampbell-msft:


package.json
```diff
@@ -849,7 +849,6 @@
     }
   },
   "scripts": {
-    "vscode:prepublish": "yarn compile-production",
```
benmcmorran, line 852: Will this break external contributors trying to package the VSIX themselves?

benmcmorran (approved):

PR #655, opened 2024-09-20
update build clean command wording

gcampbell-msft:
There was customer feedback that the wording of "Build clean the current target" and "Build clean the target all" was confusing.

This updates to be "Clean and build the current target" and "Clean and build the target ALL".


snehara99:
Is the target ALL different from all targets? Might be clearer to say that if so

gcampbell-msft:
> Is the target ALL different from all targets? Might be clearer to say that if not

So, my understanding is that the target "ALL" is a target that will build all targets needed to complete a build.

gcampbell-msft:
@snehara99 Ping for review. Thanks

snehara99 (approved):

Issue #526, opened 2023-11-13
The "#define"  information in the Makefile will be recognized as unrecognized token


Ivanbeethoven:
This example comes from AFL: https://github.com/google/AFL

This "#define" does not exist in the C code, but is passed in through the compiler command in the Makefile file. 
An error is still reported after being recognized by the makefile.
It seems that define was indeed recognized and expanded, but an error occurred

![image](https://github.com/microsoft/vscode-makefile-tools/assets/65935895/94be3118-de60-49b6-83cb-5d040750facf)

Here is my `.vscode/c_cpp_properties.json`: 
```json
{
    "configurations": [
        {
            "name": "Linux",
            "includePath": [
                "${workspaceFolder}/**"
            ],
            "defines": [],
            "compilerPath": "/usr/bin/gcc",
            "cStandard": "c11",
            "intelliSenseMode": "linux-gcc-x64",
            "configurationProvider": "ms-vscode.makefile-tools"
        }
    ],
    "version": 4
}
```
I have temporarily made the following modifications to alleviate this error:

```json
{
      "configurations": [
        {
          "name": "Linux",
          "includePath": [
            "${workspaceFolder}/**"
          ],
          "defines": [],
          "compilerPath": "/usr/bin/gcc",
          "cStandard": "c99",
          "intelliSenseMode": "linux-gcc-x64",
          "compilerArgs": [
              "-O3",
              "-funroll-loops",
              "-Wall",
              "-D_FORTIFY_SOURCE=2",
              "-g",
              "-Wno-pointer-sign",
              "-DAFL_PATH=\"$(HELPER_PATH)\"",
              "-DDOC_PATH=\"$(DOC_PATH)\"",
              "-DBIN_PATH=\"$(BIN_PATH)\""
            ]
          }
      ],
      "version": 4
}
```

gcampbell-msft:
@Ivanbeethoven Thank you for bringing this to our attention and pointing us to the project that it's happening in!

Since there is a reasonable workaround, this might not get fixed immediately (in the next release), but we will definitely put it on the backlog to investigate a fix for an upcoming release. Thanks!

snehara99:
I believe this has to do with: https://github.com/microsoft/vscode-makefile-tools/issues/465 and https://github.com/microsoft/vscode-makefile-tools/issues/497

snehara99:
@Ivanbeethoven another temporary workaround that I tested and works is to use...
[...]

Issue #352, opened 2022-08-10
Add support for cpptools-api v6 - Use compilerFragments instead of compilerArgs


Colengms:
In cpptools-api v6, a new field was added (`compilerFragments`) to disambiguate whether shell quoting/escaping is expected or not in compiler arguments.  `compilerFragments` is used specifically for command line fragments (potentially containing shell quoting and/or escaping), with `compilerArgs` intended for compiler arguments that do not contain any shell quoting or escaping.

In the case of makefile tools, that should remove the necessity of having to parse the command line into separate arguments to send in `compilerArgs`.  `compilerFragments` can accept multiple arguments in a single entry, and cpptools will parse them as the shell would to divide them into arguments.

PR #675, opened 2024-12-09
Fix "Unrecognized Token" Intellisense Issue with "Define" Flag

snehara99:
This is for issue #526 

There was an issue parsing define cflags in the form of -DFLAG= vs -D 'FLAG=...'. Now both work, but this solution should be replaced [once compiler fragments are fully adopted in makefile-tools](https://github.com/microsoft/vscode-makefile-tools/issues/352).


src/parser.ts
```diff
@@ -1208,6 +1212,21 @@ export async function parseCustomConfigProvider(
         // Exclude switches that are being processed separately (I, FI, include, D, std)
         // and switches that don't affect IntelliSense but are causing errors.
         let compilerArgs: string[] = [];
+        let compilerFragments: string[] = [];
+
+        if (useCompilerFragments) {
+          // This is a temporary solution where we are only using compiler fragments here to pass the
```
snehara99, line 1218: @gcampbell-msft let me know what you think about this

PR #685, opened 2025-01-10
[Auto] Localization - Translated Strings

csigs:


gcampbell-msft:
Just loc, won't affect build.

PR #689, opened 2025-02-13
run cg weekly

gcampbell-msft:


gcampbell-msft:
No need for tests, merging.

Issue #716, opened 2025-03-24
automationProfile ignores args Parameter


koneu:
### Brief Issue Summary

I tried setting up the extension with msys2.

Im using both `terminal.integrated.profiles.windows` and `terminal.integrated.automationProfile.windows` with identical values.

When compiling via terminal compilation succeeds, when configuring the extension i get a `make.exe: command not found` error indicating, that the `--login` option has been ignored and only the `path` is used.

___
from my settings.json:

    "terminal.integrated.automationProfile.windows": {
        "path": "C:/msys64/usr/bin/bash.exe",
        "args": [
            "--login",
            "-i",
        ],
        "env": {
            "MSYSTEM": "UCRT64",
            "CHERE_INVOKING": "1",
            "MSYS2_PATH_TYPE": "inherit",
        },
    },

### Debug Log

```shell
dryrun.log and targets.log:

make.exe --dry-run --keep-going --print-directory
/usr/bin/bash: line 1: make.exe: command not found

___
```

Amy-Li03:
Hi @koneu , thanks for reporting issue here!
Verified this on Makefile Tool v0.13.1(pre-release), unfortunately we cannot repro your issue, configure succeeded in extension, please see following screenshot.

And you can try setting the environment for make to be recognized by VS Code by adding MSYS2 to your PATH.

![Image](https://github.com/user-attachments/assets/93f83f59-2db3-4149-af4c-28d657784ab3)

koneu:
hi and thanks for the reply,

when using a login/ interactive shell the msys-startup files should get executed (bash_profile bshrc etc.). 

In a default setup this should make adding make to the path not necessary, even if its not inthe PATH on the host machine. While adding it to the PATh on the Machine would be a workaround in this case; it would not work on more complicated Startup scripts.

gcampbell-msft:
@koneu Thanks for the issue, we have known issues and/or improvements needed when it comes to the automationProfile settings in VS Code. I'll mark this accordingly and add it to our backlog.

PR #721, opened 2025-04-02
Automation profile include args

koneu:
fixes https://github.com/microsoft/vscode-makefile-tools/issues/716

i don't usually code in typeskript so i don't know how usefull this is.
also i could only test on a windows -> msys setup.

The fix is somewhat convoluted, since child_process.spawn does not appear to support specifying shell arguments.

i also didn't let 'cmd' / '/bin/bash' commands execute via the specified shell, because escaping the caracters correctly seems impossible; this affects the call from parser.ts

the setup also only works if gcc is in PATH, but that seems to be because c_cpp extension does not use the provided shell to lookup system path from the given compiler.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -399,43 +399,79 @@ export function spawnChildProcess(
         `integrated.automationShell.${shellPlatform}`
       ); // and replaced with automationProfile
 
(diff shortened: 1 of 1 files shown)


koneu:

PR #727, opened 2025-04-08
[Auto] Localization - Translated Strings

gcampbell-msft:


gcampbell-msft:
@gcampbell-msft This PR shows that the format I tried to add comments to our localized files was incorrect, I'll make an update in a subsequent PR.

snehara99 (approved):

PR #728, opened 2025-04-11
Fix loc comments by using correct format

gcampbell-msft:
See title.


snehara99 (approved):
wow this must have been an annoying update haha

PR #729, opened 2025-04-13
Fix invalid parsing of quoted argument values

ashenBlade:
Currently, in `src/parser.ts` for parsing cmd arguments from Makefile regexp is used.
But it does not handle cases when some argument have assignment with quoted values.

I have encountered this error, when working with PostgreSQL build arguments. Given this simple example:

```bash
-DVAL_CFLAGS="\\"-Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wshadow=compatible-local -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -g -O0 -g \\""
```

Regexp parse only this part `-DVAL_CFLAGS="\\"-Wall` and other will become independent.
That's ok for `-Wmissing-prototypes` or `-g`, but eventually it encounter last `\\""` and _just add `\\""` to plain list of arguments_, so this become invalid bash string and I get error like:

```
/bin/sh: 1: Syntax error: Unterminated quoted string
```

This patch fixes this by handing such quoted strings - but only them, meaning it does not handle independent quoted strings without some assignment.
Implemented by adding `(=(\'|\"))?` to regexp pattern, searching for string end and updating parsing state.

Also, I have found error in formatting error message, when parsing ends with error - no return code passed for format string, so only 2 args are passed to 3-arged format string.


ashenBlade:

gcampbell-msft:
@ashenBlade Thanks for the contribution! Could you update the CHANGELOG as well? Feel free to give yourself credit for the contribution following the pattern already in the CHANGELOG

ashenBlade:
Hi! Thanks for response. Changelog is updated.

gcampbell-msft (approved):

PR #738, opened 2025-06-19
Support mingw-make's dryrun output format

vannaka:
Add support for single quotes around the tool invocation within the dryrun output.

This is needed to support make variants which use single quotes instead of double quotes, such an mingw-make.


vannaka:
Fixes https://github.com/microsoft/vscode-makefile-tools/issues/739

vannaka:

snehara99:
@vannaka Do you think you could resolve the conflicts for me? Also, if you'd like you can give yourself credit in the changelog like this contributor does [here](https://github.com/vannaka/vscode-makefile-tools/blob/7f60bd919471b6e210692fb7543c0be550b94278/CHANGELOG.md?plain=1#L113).

vannaka:
@snehara99 Done!

snehara99 (approved):

snehara99:
@vannaka Thanks for your contribution! I'm glad we could work this out :)

PR #746, opened 2025-07-29
Add an api implementation

gcampbell-msft:
This adds an API implementation so that other vscode extensions can use an official api to interact with this extension. 

The API has methods for listing the build targets, building, and cleaning.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1403,7 +1403,7 @@ export function createCombinedCancellationToken(...tokens: (vscode.CancellationT
     }
 
     combinedSource.token.onCancellationRequested(() => {
-        disposables.forEach(d => d.dispose());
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -911,7 +911,8 @@
     "vscode-cpptools": "^6.1.0",
     "vscode-nls": "^5.0.0",
     "@vscode/extension-telemetry": "^0.9.6",
-    "vscode-jsonrpc": "^3.6.2"
+    "vscode-jsonrpc": "^3.6.2",
+    "vscode-makefile-tools-api": "file:../vscode-makefile-tools-api"
```
gcampbell-msft, line 915: build and tests will fail until this is updated after api is published

PR #809, opened 2026-04-24
fix name handling in launch config

### LINKED ISSUE #808: [Bug] when a launch configuration has a name field, the project outline always shows [unset] despite being properly set
### Brief Issue Summary

When a makefile.launchConfigurations entry includes the name field (added in #793), the Makefile: Project
Outline panel always displays the launch target as "Unset" after selection, even though the configuration
is correctly matched internally.

Root Cause

LaunchTargetNode.getShortLaunchTargetName() in tree.ts receives the display name (e.g., "Run with
arguments") from launchConfigurationDisplayName(). It passes this to stringToLaunchConfiguration(), which
expects the technical cwd>binary(args) format. Since a user-friendly name doesn't match that format,
parsing returns undefined, and the tree unconditionally falls back to showing "Unset".

A secondary issue is that readCurrentLaunchConfiguration() in configuration.ts passes the prose string "No
launch configuration set." to updateLaunchTarget() when no configuration is active. This non-empty string
also fails parsing but isn't truly a display name — it leaks status-bar text into the tree.

Steps to Reproduce

 1. Add a launch configuration with a name field in .vscode/settings.json: 
"makefile.launchConfigurations": [
      {
          "name": "Run with arguments",
          "cwd": "${workspaceFolder}",
          "binaryPath": "${workspaceFolder}/out.exe",
          "binaryArgs": ["arg1", "arg2"]
      }
  ]
 2. Run Makefile: Configure
 3. Run Makefile: Set the make launch configuration and select "Run with arguments"
 4. Observe the Makefile: Project Outline panel

Expected: Launch target shows Launch target: [Run with arguments] Actual: Launch target shows Launch
target: [Unset]

The Makefile Tools output panel confirms the configuration was matched successfully (Reading current
launch configuration "Run with arguments" from the workspace state.), so this is purely a display issue in
the tree view.

### Debug Log

```shell
n/a
```

### Additional Information

_No response_

snehara99:
Problem

PR #793 added name and description fields to makefile.launchConfigurations, but introduced a regression
where the Project Outline panel displays incorrect launch target information. Fixes #808.

Changes

1. Fix named configs showing "Unset" in Project Outline (tree.ts)

getShortLaunchTargetName() receives the display name (e.g., "Run with arguments") from 
launchConfigurationDisplayName(), but passes it to stringToLaunchConfiguration() which expects the
technical cwd>binary(args) format. Parsing fails and the tree unconditionally falls back to "Unset".

Fix: When stringToLaunchConfiguration() fails, use the input as-is if non-empty — it's a user-friendly
display name. Only fall back to "Unset" when the input is empty.

2. Fix display name lost on restart (configuration.ts)

After readCurrentLaunchConfiguration() correctly updates the tree with the display name, 
initFromSettings() overwrites it by passing extension.getState().launchConfiguration (the raw technical
string) to _projectOutlineProvider.update().

Fix: Pass launchConfigurationDisplayName(currentLaunchConfiguration) instead of the raw state string, so
the tree always receives the friendly name when one exists.

3. Fix unnamed configs missing args in Project Outline (tree.ts)

getShortLaunchTargetName() only showed the binary path (e.g., out.exe) for unnamed configs, dropping the
args portion entirely.

Fix: Include args in the short name so unnamed configs display as out.exe(-flag) instead of just out.exe.

4. Fix prose string leaking into tree contract (configuration.ts)

readCurrentLaunchConfiguration() passed the prose string "No launch configuration set." to 
updateLaunchTarget() when no configuration was active, leaking status-bar text into the tree.

Fix: Pass "" instead, with the status bar receiving the prose string via a separate fallback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Bug Fixes:
 - Fix "Makefile Tools" being translated in localized languages when it should remain as a proper noun. [#731](https://github.com/microsoft/vscode-makefile-tools/issues/731)
 - Fix binary path links in output log not being clickable on Windows when the path has no .exe extension. [#698](https://github.com/microsoft/vscode-makefile-tools/issues/698)
 - Fix `parseCompilerArgs.bat` failing when macro definitions contain numeric expressions like `-1` inside quoted values. [#755](https://github.com/microsoft/vscode-makefile-tools/issues/755)
+- Fix Project Outline showing "Unset" for launch configurations that have a `name` field. [#808](https://github.com/microsoft/vscode-makefile-tools/issues/808)
 
 ## 0.11
 
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -932,7 +932,7 @@ async function readCurrentLaunchConfiguration(): Promise<void> {
     );
   }
 
-  let launchConfigStr: string = "No launch configuration set.";
+  let launchConfigStr: string = "";
   if (currentLaunchConfiguration) {
     launchConfigStr = launchConfigurationDisplayName(currentLaunchConfiguration);
     logger.message(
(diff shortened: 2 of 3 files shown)


snehara99 (commented):
validated locally

moyo1997 (approved):

Issue #397, opened 2022-12-12
A broken makefile that recurses kills server


the-moog:
If you open a folder remotely that contains a broken makefile which can recurse infinately then simply opening a remote folder kills the server before you have a chance to stop it.
Suggestion:
If Makefile tools fails to initialise on a run then suspend the tool for that folder until user action corrects the error.

andreeis:
@the-moog, thank you for the feature suggestion. When that happens, you do not see on the bottom right a popup with details about which phase of the "configure" process we're in and a "cancel" button? We may have a different bug that we'd like to investigate and address before implementing the idea that you suggest.
Can you confirm (and attach a screenshot) with what happens when you open your "broken" makefile project?

the-moog:
Hi @andreeis .  Not easy for me to reproduce as I don't recall what project/folder had the recursion, what versions (vscode/make-tools etc) and I will have since corrected the error in the Makefile (a simple use of $@ instead of a  $< in a recipe)

I created a new .vscode folder and a two line recursive Makefile to see what would happend and I did not get the same result.  Though I am not sure I am even installing the same Makefile-tools version in that remote folder, and due to the auto-upgrade of vscode may even be using a different version. Perhaps there are other aspects of the specific Makefile that cause this behaviour.

Amongst a lot more, I know the Makefile will use the SHELL= and .SUFFIXES: stanzas.  I tried those on this Makefile with no change in behaviour.

I do note that I don't recognise much of what is in the log compared to what I saw when it went wrong.  I think it only got as far as "dry-run", whereas this time it went past that (even though dry run failed)

I can only say for sure that the dialog 'Configuring Extension makefile-tools' remains stuck on the screen. It can be dismissed but if you look at the console it would never complete.

I guess perhps this is fixed?  If it happens again I will try and make a better report.

----


recursive_make/Makefile
```make
target.file:
	$(MAKE) target.file
```

Log from 
.vscode-server/data
```
Logging level: Normal
Dropping various extension output files at /home/jm48/.vscode-server/data/User/workspaceStorage/cf746d9ca78f510961e7da05b9f1e0bf/ms-vscode.makefile-tools
Configurations cached at /home/jm48/.vscode-server/data/User/workspaceStorage/cf746d9ca78f510961e7da05b9f1e0bf/ms-vscode.makefile-tools/configurationCache.log
No path to the makefile is defined in the settings file.
No folder path to the makefile is defined in the settings file.
Always pre-configure: false
[...]

PR #836, opened 2026-09-17
Spawn child processes in their own process group on POSIX (fixes #835)

### LINKED ISSUE #835: Dry-run during configure executes workspace recipes and can terminate VS Code (kill 0 in a Makefile trap)
## Summary

Opening a folder whose `Makefile` contains a `trap '...' EXIT` with `kill 0` terminates the entire VS Code application, before the user can interact with the window.

Two separate problems combine to cause this:

1. The configure step's `make --dry-run` is treated as inert. It isn't — GNU make deliberately **executes** recipe lines containing `$(MAKE)` even under `-n`.
2. That `make` is spawned in VS Code's own process group, so anything the workspace's Makefile signals hits the editor.

The result is that merely opening a folder lets that folder's contents kill VS Code.

## Repro

Minimal `Makefile`, no `.vscode` config needed:

```make
dev:
	@trap 'kill 0' EXIT; $(MAKE) sub & wait

sub:
	@true
```

Open the containing folder in VS Code. Makefile Tools activates, runs its configure dry-run, and the application dies.

The same thing from a terminal, isolated into its own session so it only kills itself:

```
$ perl -e 'use POSIX; if(fork()==0){ POSIX::setsid();
    exec("make --dry-run --always-make --keep-going --print-directory > out.txt 2>&1; echo EXIT=\$? >> out.txt"); } wait;'
$ cat out.txt
make: Entering directory `/tmp/repro'
trap 'kill 0' EXIT; /usr/bin/make sub & wait
make[1]: Entering directory `/tmp/repro'     <-- sub-make actually ran under --dry-run
true                                         <-- and so did its recipe
make[1]: Leaving directory `/tmp/repro'
```

Note there is no `EXIT=` line: the shell was killed by the signal before it could report a status. Under VS Code that signal goes to VS Code's process group instead.

## Why `--dry-run` doesn't protect you

From the GNU make manual, on `-n`: lines in the recipe that contain `$(MAKE)` or `${MAKE}` (and lines prefixed with `+`) are run regardless, so that recursive builds can be traced. This is intended behavior, not a make bug — but it means `--dry-run` is not a sandbox, and the configure step cannot assume it won't execute workspace code.

`--always-make`, which the extension also passes, widens this: every such line is considered out of date and runs.

## Actual behavior

Entire VS Code application exits. No error, no crash dialog, no `DiagnosticReports` entry (it's a clean SIGTERM, not a crash). `main.log` shows no failure — from the logs it looks like a normal shutdown, which makes this very hard to diagnose. The only trace is the command in `ms-vscode.makefile-tools/dryrun.log` in the workspace storage folder.

## Expected behavior

Opening a folder should never allow that folder's contents to terminate the editor.

## Suggested fix

Spawn the configure/dry-run child **detached, in its own process group/session** (`detached: true` in `child_process.spawn`, and don't `unref` it if you need to await it). Then `kill 0` from a workspace recipe can only reach the dry-run's own subtree, which is the correct blast radius. This also bounds the damage from #397 (recursive Makefile exhausting process resources) and any other recipe the dry-run ends up executing.

Secondarily, it would be worth documenting that the dry-run may execute recipe lines, since "we only dry-run it" is a reasonable thing for users to assume and it isn't true.

## Related

- #397 — "A broken makefile that recurses kills server". Same class (dry-run executing recipes takes down the host), different mechanism, and it stalled because the reporter couldn't reproduce it. This one reproduces deterministically in 4 lines.

## Environment

- VS Code: 1.139.0-insider (`046944034292b5479b4e9a50ad1a508033ffb64f`), arm64
- Makefile Tools: 0.12.17
- OS: macOS 26.6.2, Apple Silicon
- GNU Make 3.81 (Apple Command Line Tools)

Not macOS-specific — the process-group behavior is the same on Linux.

JoakimCarlsson:
Fixes #835.

## Problem

`spawnChildProcess` runs commands that are under workspace control — the configure step runs the workspace's own makefile. `--dry-run` does not make that safe: GNU make deliberately **executes** recipe lines containing `$(MAKE)` under `-n`, so configuring a folder can run arbitrary recipes.

Because the child was spawned into the extension host's process group, a recipe that signals its own group reached VS Code itself. `trap 'kill 0' EXIT` is a common way to clean up background jobs, and it's enough to terminate the entire application on folder open, before the user can interact with the window. There's no crash dialog and nothing in `main.log` — it's a clean SIGTERM, so it looks like a normal shutdown, which makes it quite hard to diagnose.

## Change

Spawn with `detached: true` on non-Windows, so the child leads its own process group and session and such a signal is bounded to the spawned subtree.

Deliberately **not** applied on Windows, where `detached` allocates a new console window instead, and where `killTree` already isolates via `taskkill`.

The child is **not** `unref`'d, so the promise still awaits it and stdio callbacks are unchanged. `killTree` is unaffected: it walks children by PID via `pgrep -P`, not by process group.

## Testing

**1. Mechanism** — harness replicating this call site's exact options (`shell: true`, same env), running a makefile with `trap 'kill 0' EXIT`, each variant in its own session so it only kills itself:

| | `detached: false` (before) | `detached: true` (after) |
|---|---|---|
| child | killed | killed by SIGTERM (its own group) |
| **parent** | **killed, never reported** | **survived, observed `signal=SIGTERM`** |

The parent stands in for the extension host. Note the fix also preserves the caller's ability to observe the signal.

**2. End to end** — packaged a VSIX from this branch, installed it in VS Code Insiders 1.139.0-insider (macOS 26.6.2, arm64, GNU Make 3.81), and opened a trusted folder containing:

```make
dev:
	@trap 'kill 0' EXIT; $(MAKE) sub & wait

sub:
	@true
```

With the shipped 0.12.17 this terminates VS Code. With this change the editor stays up, and the configure completes normally — `dryrun.log` shows the recipe really did run and the trap really did fire:

```
make --dry-run --always-make --keep-going --print-directory
make: Entering directory `.../mftest-detach'
trap 'kill 0' EXIT; /Library/Developer/CommandLineTools/usr/bin/make sub & wait
make[1]: Entering directory `.../mftest-detach'
true
make[1]: Leaving directory `.../mftest-detach'
```

`targets.log` and `configurationCache.log` were produced as usual, so configure is not degraded — only the blast radius changed.

`yarn compile` is clean. I checked `src/util.ts` against Prettier before and after; the only differences introduced are the added lines, so I left the file's pre-existing formatting alone rather than widening the diff.

## Notes

- I did not touch `package.json` or `yarn.lock`, per CONTRIBUTING.
- This also bounds the damage from #397, where a recursive makefile exhausts process resources during the same dry-run — that one isn't fixed here, but it can no longer take the host's process group with it.
- Worth considering separately: the configure step executing workspace recipes at all is surprising, and "we only dry-run it" is a reasonable assumption for users to make. A note in the docs may be warranted.


JoakimCarlsson:

